PR #3, opened 2021-03-02
ci(actions): add reuse compliance workflow

s4heid:


kristian:
would you like to add a small description to the yaml file, similar to our other GitHub actions, to explain what this workflow is supposed to be doing?

kristian (approved):

PR #8, opened 2021-03-05
refactor: simplify the EntityCollectionProcessor

pk-work:
This patch ...

- ... renames getEntities(...) to applyFilterQueryOption(...) because
getEntities(...) was misleading and inconsistent.

- ... removes isDebugEnabled(), because nothing was computed inside.

- ... changes all apply...QueryOption method signatures to throw an
error instead of failing the passed processPromise. This is required,
because if the processPromise has been failed the processing of the
request must be stopped.

- ... removes the count logic from readEntityCollection(...), because it
will be processed in countEntityCollection(...).

- ... removes applySelectQueryOption(...), because it only builds a
select list which should be done at a more central place. This change is
also a prerequisite for the expand query option, which also needs to
modify the select list.

- ... simplifies the countEntityCollection(...) method dramatically.


src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
```diff
@@ -86,87 +83,91 @@ public void readEntityCollection(ODataRequest request, ODataResponse response, U
...
+                }
+
+                EntityCollectionSerializerOptions opts = createSerializerOptions(request, uriInfo, edmEntitySet);
+                response.setContent(odata.createSerializer(responseFormat)
+                        .entityCollection(serviceMetadata, edmEntityType, entityCollection, opts).getContent());
+                response.setStatusCode(HttpStatusCode.OK.getStatusCode());
+                response.setHeader(HttpHeader.CONTENT_TYPE, responseFormat.toContentTypeString());
+                getProcessPromise().complete();
+            } catch (ODataException e) {
+                getProcessPromise().fail(e);
```
kristian, line 106: Same here

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
```diff
@@ -284,42 +226,36 @@ public void countEntityCollection(ODataRequest request, ODataResponse response,
...
+                 * with the path segment /$count.
+                 */
+                List<Entity> resultEntityList = applyFilterQueryOption(uriInfo.getFilterOption(), ew.getEntities());
+
+                ByteArrayInputStream serializerContent = new ByteArrayInputStream(
+                        String.valueOf(resultEntityList.size()).getBytes(StandardCharsets.UTF_8));
+                response.setContent(serializerContent);
+                response.setHeader(HttpHeader.CONTENT_TYPE, ContentType.TEXT_PLAIN.toContentTypeString());
+                response.setStatusCode(HttpStatusCode.OK.getStatusCode());
+                getProcessPromise().complete();
```
kristian, line 246: Same here

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
```diff
@@ -284,42 +226,36 @@ public void countEntityCollection(ODataRequest request, ODataResponse response,
...
+            } catch (ODataException e) {
+                getProcessPromise().fail(e);
```
kristian, line 248: Same here

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
```diff
@@ -86,87 +83,91 @@ public void readEntityCollection(ODataRequest request, ODataResponse response, U
...
+                    resultEntityList = applyTopQueryOption(uriInfo.getTopOption(), resultEntityList);
+                    entityCollection.getEntities().addAll(resultEntityList);
+                }
+
+                EntityCollectionSerializerOptions opts = createSerializerOptions(request, uriInfo, edmEntitySet);
+                response.setContent(odata.createSerializer(responseFormat)
+                        .entityCollection(serviceMetadata, edmEntityType, entityCollection, opts).getContent());
+                response.setStatusCode(HttpStatusCode.OK.getStatusCode());
+                response.setHeader(HttpHeader.CONTENT_TYPE, responseFormat.toContentTypeString());
+                getProcessPromise().complete();
```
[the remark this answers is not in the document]
pk-work: I think it is only called once, because if it passes the try block, the catch block isn't evaluated.

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
pk-work: I think it is only called once, because if it passes the try block, the catch block isn't evaluated.

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
pk-work: I think it is only called once, because if it passes the try block, the catch block isn't evaluated.

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
pk-work: I think it is only called once, because if it passes the try block, the catch block isn't evaluated.

src/main/java/io/neonbee/internal/processor/EntityCollectionProcessor.java
```diff
@@ -86,87 +83,91 @@ public void readEntityCollection(ODataRequest request, ODataResponse response, U
          */
         EdmEntitySet edmEntitySet = ((UriResourceEntitySet) uriInfo.getUriResourceParts().get(0)).getEntitySet();
         EdmEntityType edmEntityType = edmEntitySet.getEntityType();
+
         // Fetch the data from backend
-        Promise<Void> processPromise = getProcessPromise();
+        fetchEntities(request, edmEntityType, ew -> {
+            try {
+                List<Entity> resultEntityList = applyFilterQueryOption(uriInfo.getFilterOption(), ew.getEntities());
```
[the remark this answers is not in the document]
pk-work: Done

PR #9, opened 2021-03-08
build: add required configuration for publishing to maven central

pk-work:
This patch ...

* ... adds meta information for license, scm or developers to the
pom.xml which is required by maven central.

* adds a task to create a test Javadoc artifact, which is required by
maven central.

* adds fixes for violations in the test Javadoc

* adds the script 'publishToMaven.sh' which can be used to add required
credentials when publishing to maven central.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,8 +22,7 @@ plugins {
 }
 
 group = 'io.neonbee'
-// have to use single quotes for XMake, change as soon as new version Gradle build plugin is merged
-version = '0.0.8'
+version = '0.2.2'
 mainClassName = 'io.neonbee.Launcher'
 archivesBaseName = 'neonbee-core'
 sourceCompatibility = 11
--- a/gradle/publishing.gradle
+++ b/gradle/publishing.gradle
@@ -1,10 +1,7 @@
 apply plugin: 'maven-publish'
+apply plugin: 'signing'
 
 publishing {
-    repositories {
-        mavenLocal()
-    }
-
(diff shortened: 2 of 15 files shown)


build.gradle
```diff
@@ -23,7 +23,7 @@ plugins {
 
 group = 'io.neonbee'
 // have to use single quotes for XMake, change as soon as new version Gradle build plugin is merged
```
kristian, line 25: shall we remove all references to "XMake" in the comments? Also maybe search for it globally

gradle/publishing.gradle
```diff
@@ -13,6 +10,44 @@ publishing {
             artifact source: shadowJar
             artifact source: sourcesJar
             artifact source: javadocJar
+
+            pom {
+                name = 'NeonBee Core'
+                description = 'NeonBee is an open source reactive dataflow engine, a data stream processing framework using Vert.x.'
+                url = 'https://neonbee.io/'
+
+                licenses {
```
kristian, line 19: I saw this information is redundant in the files quite often, isn't there a way to specify this only once?

kristian (changes requested):

build.gradle
pk-work: fixed

gradle/publishing.gradle
pk-work: I don't know how. :(

s4heid (approved):

PR #15, opened 2021-03-12
refactor: Simplify EntityProcessor

pk-work:
This patch ...

... simplifies the EntityProcessor by replacing ambiguous method names
and logic.

... removes the EntityProcessorTest, because all the tests are already
part of ODataUpdateEntityTest, ODataCreateEntityTest,
ODataDeleteEntityTest and ODataReadEntityTest.

... improves error handling in ODataEndpointHandler, by setting the
correct status code in case that the error is of type
ODataApplicationException.


src/main/java/io/neonbee/internal/processor/EntityProcessor.java
```diff
@@ -90,90 +92,99 @@ public void init(OData odata, ServiceMetadata serviceMetadata) {
...
-        }
-        handleEntity(edmEntityType, request, action, entity, entityHandler);
+        EdmEntityType entityType = uriResourceEntitySet.getEntitySet().getEntityType();
+        return odata.createDeserializer(requestFormat).entity(request.getBody(), entityType).getEntity();
     }
 
-    void handleEntity(EdmEntityType edmEntityType, ODataRequest request, DataAction action, Entity entity,
-            Handler<AsyncResult<EntityWrapper>> entityHandler) {
+    @VisibleForTesting
+    Future<EntityWrapper> forwardRequest(ODataRequest request, UriResourceEntitySet uriResourceEntitySet, Entity entity,
```
SAPDaniloWork, line 164: i guess this method can be declared as private and remove visible for testing pls

src/main/java/io/neonbee/internal/processor/EntityProcessor.java
```diff
@@ -90,90 +92,99 @@ public void init(OData odata, ServiceMetadata serviceMetadata) {
...
 
-        requestEntity(vertx, new DataRequest(edmEntityType.getFullQualifiedName(), query),
-                new DataContextImpl(routingContext)).onComplete(entityHandler);
+        return requestEntity(vertx, new DataRequest(entityType.getFullQualifiedName(), query),
+                new DataContextImpl(routingContext)).onFailure(processPromise::fail);
     }
 
     @VisibleForTesting
     @SuppressWarnings("PMD.PrematureDeclaration")
-    Handler<AsyncResult<EntityWrapper>> handleReadEntityResult(UriResourceEntitySet uriResourceEntitySet,
```
SAPDaniloWork, line 167: i guess this method can be declared as private and remove visible for testing pls

src/main/java/io/neonbee/internal/processor/EntityProcessor.java
```diff
@@ -90,90 +92,99 @@ public void init(OData odata, ServiceMetadata serviceMetadata) {
...
-            EdmEntitySet edmEntitySet, UriInfo uriInfo, ODataResponse response, ContentType responseFormat,
-            HttpStatusCode responseStatus) {
-        return asyncEntity -> handleResult(asyncEntity, processPromise -> {
+    Handler<EntityWrapper> handleReadEntityResult(UriResourceEntitySet uriResourceEntitySet, UriInfo uriInfo,
+            ODataResponse response, ContentType responseFormat, HttpStatusCode responseStatus,
+            Promise<Void> processPromise) {
+        return ew -> {
             // The result entity that will be returned
             Entity foundEntity = null;
```
SAPDaniloWork, line 183: why put this not below where the variable will actually used at line 190 (above the first try catch)?

src/main/java/io/neonbee/internal/processor/EntityProcessor.java
```diff
@@ -88,148 +88,119 @@ public void init(OData odata, ServiceMetadata serviceMetadata) {
...
-                } catch (SerializerException e) {
-                    processPromise.fail(
-                            new ODataApplicationException("Error during processing select option: " + e.getMessage(),
-                                    HttpStatusCode.INTERNAL_SERVER_ERROR.getStatusCode(), Locale.ENGLISH));
-                }
-            }
-
-            // Return the OData response
+    private Handler<EntityWrapper> handleReadEntityResult(UriResourceEntitySet uriResourceEntitySet, UriInfo uriInfo,
+            ODataResponse response, ContentType responseFormat, HttpStatusCode responseStatus,
```
SAPDaniloWork, line 171: This value is currently always HttpStatus.OK. Maybe it make sense to not pass it. It would reduce the signature by 1 line which is also nice. If we need to parse it in future then we can pass it but for now it does not make sense
`HttpStatusCode responseStatus,`

SAPDaniloWork (approved):

PR #16, opened 2021-03-12
fix: flaky WatchVerticleTest

pk-work:


s4heid:
I triggered the voter a couple of times and it never failed. Seems like increasing the timeout actually reduces the flakyness.

s4heid (approved):

PR #29, opened 2021-03-28
fix: switch to async. log appender

kristian:
The default configuration for logback currently is using blocking appenders which, in Vert.x, is generally discouraged. Switching to logbacks AsyncAppender interface, for unblocking the logging threads.


s4heid:
closing/reopening in favour of the ci.

pk-work (approved):

PR #32, opened 2021-03-30
Make neonbee launcher options configurable via environment variables

s4heid:
For deployment on cloud environments, it is beneficial to provide the
launcher options via environment variables instead of command line
arguments. This change supports reading options either from command line
or environment variable.

### DIFF
--- a/src/test/java/io/neonbee/LauncherTest.java
+++ b/src/test/java/io/neonbee/LauncherTest.java
@@ -4,11 +4,13 @@ import static com.google.common.truth.Truth.assertThat;
 import static io.neonbee.Launcher.INTERFACE;
 import static io.neonbee.Launcher.parseOptions;
 import static io.neonbee.test.helper.FileSystemHelper.createTempDirectory;
+import static io.neonbee.test.helper.SystemHelper.withEnvironment;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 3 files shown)


src/test/java/io/neonbee/LauncherTest.java
```diff
@@ -46,26 +51,18 @@ void throwErrorIfWorkingDirValueIsEmpty() {
     }
 
     @Test
-    @DisplayName("should throw error, if instance-name is empty")
+    @DisplayName("should throw an error, if instance-name is empty")
     void throwErrorIfInstanceNameIsEmpty() {
-        args = new String[] { "-cwd", tempDirPath.toAbsolutePath().toString(), "-name", "" };
+        args = new String[] { "-cwd", workDir, "-name", "" };
         IllegalArgumentException exception =
                 assertThrows(IllegalArgumentException.class, () -> parseOptions(INTERFACE.parse(List.of(args))));
```
SAPDaniloWork, line 58: nit: you could also introduce a method to do smth like this to avoid the line break since this appears 6-7 times:
```
    @Test
    @DisplayName("should throw an error, if instance-name is empty")
    void throwErrorIfInstanceNameIsEmpty() {
        args = new String[] { "-cwd", workDir, "-name", "" };
        IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, this::parseOptionsArgs);
        assertThat(exception.getMessage()).isEqualTo("instanceName must not be empty");
    }

```

SAPDaniloWork (commented):
looks good

src/test/java/io/neonbee/LauncherTest.java
s4heid: I guess this won't work, as `parseOptionsArgs` takes an argument.

SAPDaniloWork: it works, look where args is defined

SAPDaniloWork (approved):

PR #33, opened 2021-03-30
fix: add fake URI path to navigation properties requests

pk-work:
This is required if the related EntityVerticle is serving multiple
entities and needs to decide which type should be returned. This is a
workaround until inter-verticle entity exchange [1] has been improved.

[1] https://github.com/SAP/neonbee/blob/main/docs/milestones/E1_MS02_entity_exchange.md


SAPDaniloWork (commented):
looks good

SAPDaniloWork (approved):

PR #36, opened 2021-04-06
fix: re-enable system query count

pk-work:
The method countEntityCollection of CountEntityCollectionProcessor is
only called for requests that ends with /$count. In case that count is
part of a system query, it must be implemented in the regular
readEntityCollection method.

### DIFF
--- a/src/test/java/io/neonbee/test/endpoint/odata/ODataReadEntitiesTest.java
+++ b/src/test/java/io/neonbee/test/endpoint/odata/ODataReadEntitiesTest.java
@@ -33,11 +33,16 @@ import io.neonbee.test.base.ODataRequest;
 import io.neonbee.test.endpoint.odata.verticle.TestService1EntityVerticle;
 import io.neonbee.test.helper.WorkingDirectoryBuilder;
 import io.vertx.core.DeploymentOptions;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.client.HttpResponse;
 import io.vertx.junit5.Timeout;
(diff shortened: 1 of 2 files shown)


src/test/java/io/neonbee/test/endpoint/odata/ODataReadEntitiesTest.java
```diff
@@ -95,6 +98,24 @@ void existingEntitiesTest(VertxTestContext testContext) {
...
+        Future<HttpResponse<Buffer>> response =
+                requestOData(new ODataRequest(TEST_ENTITY_SET_FQN).setQuery(countQuery));
+        assertOData(response, body -> {
+            JsonObject json = body.toJsonObject();
+            assertThat(json.getMap()).containsAtLeast("@odata.count", 5);
+        }, testContext)
+                .compose(v -> assertODataEntitySetContainsExactly(response,
+                        List.of(EXPECTED_ENTITY_DATA_1, EXPECTED_ENTITY_DATA_2, EXPECTED_ENTITY_DATA_3,
+                                EXPECTED_ENTITY_DATA_4, EXPECTED_ENTITY_DATA_5),
+                        testContext))
```
SAPDaniloWork, line 115: avoid redundant code:
```
 List.of(EXPECTED_ENTITY_DATA_1, EXPECTED_ENTITY_DATA_2, EXPECTED_ENTITY_DATA_3,
                                EXPECTED_ENTITY_DATA_4, EXPECTED_ENTITY_DATA_5)
```
is being used twice in the tests (in this test and in "existingEntitiesTest"). If you declare it as constant you can reuse it and it would make the tests smaller.

SAPDaniloWork (dismissed):

SAPDaniloWork (approved):

PR #40, opened 2021-04-23
Pull request for completing configurable endpoints milestone

kristian:


pk-work:
You introduce a "getServerPort" method in NeonBeeOptions, but it is never used. Because the ServerVerticle is reading the "port" property from the ServerConfig. Based on our discussions about config / options, the DEFAULT value in the ServerConfig should be the port specified in the options and the default in the options should be 8080.

resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
```diff
@@ -36,32 +36,46 @@ config:
...
+      # provides a OData V4 compliant endpoint, for accessing entity verticle data
+      - type: io.neonbee.endpoint.ODataV4Endpoint
+        # enable the OData endpoint, defaults to true
+        enabled: true
+        # the base path to map this endpoint to, defaults to /odata/
+        basePath: /odata/
+        # endpoint specific authentication chain, defaults to null and using the default authentication chain
+        authenticationChain: ~
+        # namespace and service name URI mapping (strict, or loose based on CDS)
+        uriConversion: strict
```
pk-work, line 48: This is wrong, isn't it? It should be:

additionalConfig:
  uriConversion: strict

kristian: Hey @pk-work, I wanted to not overcomplicate the config file and I found it strage that you have a "additonal**Config**" in a "**config**" file. Same as I found it strage to name it "endpoints**Config**". Thus I decided to introduce new "collectAdditionalConfig" and "rephraseConfigNames" methods. However, due to that I found this issue when I fixed the unit tests, it's part of my "fix unit tests" commit and you will find it there. This change allows you to specify addtional configurations on the same level and it also allows you to simply specify "endpoints" and not "endpointConfigs" in the YAML file.

kristian:
> You introduce a "getServerPort" method in NeonBeeOptions, but it is never used. Because the ServerVerticle is reading the "port" property from the ServerConfig. Based on our discussions about config / options, the DEFAULT value in the ServerConfig should be the port specified in the options and the default in the options should be 8080.

Good catch! I think I removed:

```
int port = Optional.ofNullable(options.getServerVerticlePort())
                .orElse(config.getInteger(CONFIG_PROPERTY_PORT_KEY, DEFAULT_PORT));
```

Because I thought "hey Vert.x will read it from the ServerConfig". I think this line should be added again. Thanks Pascal.

pk-work:
Should we apply spotless on generated code? I mean it is generated code ... We could also discuss about pushing this generated code into git, because it will be generated on build time.

kristian:
> We could also discuss about pushing this generated code into git, because it will be generated on build time.

I went with the recommendation of Vert.x here, they also do commit the generated code, as the repo. should be in a executable state from code side (e.g. so also forward navigation on GitHub works), when you clone the repo. Thus I would recommend committing the generated code.

> Should we apply spotless on generated code? I mean it is generated code ...

At least w/ the spotless generation our general code formatting is respected, which makes the committed code files easier to read. Thus I decided to also apply spotless on the generated files.

pk-work:
> > We could also discuss about pushing this generated code into git, because it will be generated on build time.
> 
> I went with the recommendation of Vert.x here, they also do commit the generated code, as the repo. should be in a executable state from code side (e.g. so also forward navigation on GitHub works), when you clone the repo. Thus I would recommend committing the generated code.
> 
> > Should we apply spotless on generated code? I mean it is generated code ...
> 
> At least w/ the spotless generation our general code formatting is respected, which makes the committed code files easier to read. Thus I decided to also apply spotless on the generated files.

Okay, that's fine for me

resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
pk-work: I agree to and understood the introduction of "collectAdditionalConfig". But maybe there is a better solution instead of introducing "rephraseConfigNames". We could simply rename the variables in the DataObject or define a custom "jsonPropertyNameFormatter" in the DataObject annotation, which goes in the same direction as the "rephraseConfigNames" method.

I would prefer to evaluate "jsonPropertyNameFormatter", because then we have clean variable names and use Vert.x tools instead of inventing our own rephrasing.

kristian: Agree that this would be the better solution, but such a thing we wold need to suggest to the Vert.x CodeGen first isn't it? Which could take a lot of time. I think my solution is not very unnice, however I agree, having it in the CodeGen would be nicer. I would suggest we'd stick with my suggestion for now and then propose it to the CodeGen project as a contribution, what do you think? I think the jsonPropertyNameFormatter is an experimental feature as far as I can judge the code... I think it'll be an easy improvement as a subsequent commit, which doesn't brake anything, thus I would suggest, we go with it for now and reuse the reprhaseConfigNames later on to put it into a jsonPropertyNameFormatter.

Renaming the fields in the DataObject is no good solution either... as we would need to name the property "endpoints" and have a "getEndpoints" and "setEndpoints" method. However the method names would be misleading, as the methods would return "EndpointConfig"s and NOT "Endpoint"s. That is why the rephraseConfigNames method thought makes a lot of sense.

pk-work: Okay I talked with Julien and it is not possible at the moment to use jsonPropertyNameFormatter for this, as the formatters you can define must be one of these [1]. But Julien likes the idea and it is appreciated to contribute this feature. 

Then let's continue as you suggested.

[1] https://github.com/vert-x3/vertx-codegen/blob/730d4f707ef4eb8d6dca489614b9a5e9e0e3c182/src/main/java/io/vertx/codegen/generators/dataobjecthelper/DataObjectHelperGen.java#L349

src/main/java/io/neonbee/endpoint/metrics/MetricsEndpoint.java
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * The default path the metrics endpoint is exposed by NeonBee.
+     */
+    public static final String DEFAULT_BASE_PATH = "/metrics/";
+
+    @Override
+    public EndpointConfig getDefaultConfig() {
+        // as the EndpointConfig stays mutable, do not extract this to a static variable, but return a new object
+        return new EndpointConfig().setType(MetricsEndpoint.class.getName()).setBasePath(DEFAULT_BASE_PATH)
+                // by default the metrics endpoint does not require authentication
```
pk-work, line 24: I see this differently. Exposing metrics to everyone is in my opinion a non necessary security issue. Depending on the exposed metrics, someone could figure out what helps best to bring down a service. Or it could even motivate people to bring the service to the limit, because they see live the pressure on the system.

My suggestion: It should be secured by the default authchain by default

src/main/java/io/neonbee/endpoint/odatav4/internal/olingo/OlingoEndpointHandler.java
```diff
@@ -43,245 +31,44 @@
...
      *
      * @param serviceMetadata The metadata of the service
-     * @return A ODataEndpointHandler instance
+     * @return A OlingoEndpointHandler instance
      */
-    public static ODataEndpointHandler create(ServiceMetadata serviceMetadata) {
-        return new ODataEndpointHandler(serviceMetadata);
+    public static OlingoEndpointHandler create(ServiceMetadata serviceMetadata) {
+        return new OlingoEndpointHandler(serviceMetadata);
     }
```
pk-work: This is something a recognized already a while ago, The reason why Vert.x has these convenience methods is, that their handlers are Interfaces which offers then concrete implementations with these methods. In this case OlingoEndpointHandler is already a concrete implementation and must be compared with "ErrorHandlerImpl" which doesn't offer a static create method.

I would  like to remove the static create method here.

src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
```diff
@@ -0,0 +1,240 @@
...
+
+        /**
+         * Convenience method as similar other Vertx handler implementations (e.g. ErrorHandler)
+         *
+         * @param config the configuration of the endpoint
+         * @return A RawDataEndpointHandler instance
+         */
+        public static RawHandler create(JsonObject config) {
+            return new RawHandler(config);
+        }
```
pk-work, line 93: same here

src/main/java/io/neonbee/internal/handler/HooksHandler.java
```diff
@@ -23,15 +23,14 @@
      *
      * @return the HooksHandler
      */
-    public static Handler<RoutingContext> create() {
+    public static HooksHandler create() {
         return new HooksHandler();
     }
```
pk-work, line 28: same here

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -1,366 +1,283 @@
...
-            throw new UnsupportedOperationException("HtDigest authentication provider is not implemented yet");
-        case "htpasswd":
-            // TODO return HtpasswdAuth.create(vertx, new HtpasswdAuthOptions(jsonConfig));
-            throw new UnsupportedOperationException("HtPasswd authentication provider is not implemented yet");
-        case "jdbc":
-            // TODO return JDBCAuth.create(vertx, client).setAllSettersHere;
-            throw new UnsupportedOperationException("JDBC authentication provider is not implemented yet");
-        case "jwt":
-            JsonArray pubSecKeys = options.getJsonArray("pubSecKeys");
-            return JWTAuth.create(vertx, new JWTAuthOptions().setPubSecKeys(extractPubSecKeys(pubSecKeys)));
```
pk-work, line 311: You have never replaced that, but it is necessary. Maybe it is fixed in Vert.x 4.0.3 but I don't think so.

src/main/java/io/neonbee/config/ServerConfig.java
```diff
@@ -0,0 +1,798 @@
...
+        // we override some settings like the default port, use alpn and compression settings
+        setPort(json.getInteger(PROPERTY_PORT, DEFAULT_PORT));
+        setUseAlpn(json.getBoolean("useAlpn", DEFAULT_USE_ALPN));
+        setCompressionSupported(json.getBoolean("compressionSupported", DEFAULT_COMPRESSION_SUPPORTED));
+        setCompressionLevel(json.getInteger("compressionLevel", DEFAULT_COMPRESSION_LEVEL));
+        setDecompressionSupported(json.getBoolean("decompressionSupported", DEFAULT_DECOMPRESSION_SUPPORTED));
+    }
+
+    @Override
+    public JsonObject toJson() {
```
pk-work, line 247: I think toJson should also respect the attributes stored in super class, shouldn't it?

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -1,366 +1,283 @@
...
+
+        // add all endpoint handlers as sub-routes here
+        mountEndpoints(router, config.getEndpointConfigs(), defaultAuthHandler, hooksHandler)
+                .onFailure(startPromise::fail).onSuccess(nothing -> {
+                    // register a fallback route and the NotFoundHandler, which finally fails
+                    Route fallbackRoute = router.route();
+                    defaultAuthHandler.ifPresent(fallbackRoute::handler);
+                    fallbackRoute.handler(hooksHandler).handler(NotFoundHandler.create());
+
+                    vertx.createHttpServer(config /* ServerConfig is a HttpServerOptions subclass */)
```
pk-work, line 111: Is this the only reason why ServerConfig is extending HttpServerOptions? If yes we could maybe introduce a "toHttpServerOptions()" method. Then we would also solve the issue, that we have a lot of attributes in ServerConfig  which are not stored in the json.

pk-work: I was wrong here, they are stored

src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
kristian: Yes.

src/main/java/io/neonbee/internal/handler/HooksHandler.java
kristian: Ok.

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
kristian: What do you mean with "that"? the extractPubSecKeys? I just replaced it with the respective method in Vert.x and I checked the implementation, it looked essentially exactly like the think you did in extractPubSecKeys, or am I mistaking?

pk-work: PreProcessor shoudl replace place holder with "BEGIN PRIVATE KEY\n" + "verificationkey\n" + "END PRIVATE KEY"

PR #47, opened 2021-05-12
feat!: introduce exposedEntities/Verticles allow/block lists

pk-work:
This change introduces new properties exposedEntities for the
ODataV4Endpoint and exposedVerticles for the RawEndpoint. These settings
can be used to initialize a new RegexBlockList, which can be used to
limit the verticles / entities exposed from a given endpoint.

### DIFF
--- a/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
+++ b/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
@@ -10,7 +10,7 @@ config:
     # sets whether to use application-layer protocol negotiation or not, defaults to true
     useAlpn: true
     # one of: NONE, LOCAL or CLUSTERED, defaults to NONE
-    sessionHandling: none
+    sessionHandling: NONE
     # the name of the session cookie, defaults to neonbee-web.session
     sessionCookieName: neonbee-web.session
 
@@ -49,6 +49,11 @@ config:
         authenticationChain: ~
         # namespace and service name URI mapping (STRICT, or LOOSE based on CDS)
         uriConversion: STRICT
+        # a block / allow list of verticles to expose via this endpoint (defaults to empty / all entities exposed)
+        # the value of block / allow must be an array with Strings representing a regexp.
+        exposedEntities:
+          block: [any_allow_list_of_regexp_here]
+          allow: [any_block_list_of_regexp_here]
 
       # provides a REST endpoint (JSON, text, binary), for accessing data verticles
       - type: io.neonbee.endpoint.raw.RawEndpoint
@@ -60,6 +65,11 @@ config:
         authenticationChain: ~
         # whether or not to expose hidden verticles, defaults to false
         exposeHiddenVerticles: false
+        # a block / allow list of verticles to expose via this endpoint (defaults to empty all verticles exposed)
+        # the value of block / allow must be an array with Strings representing a regexp.
+        exposedVerticles:
+          block: [any_allow_list_of_regexp_here]
+          allow: [any_block_list_of_regexp_here]
 
       # provides an Prometheus scraping endpoint for Micrometer.io metrics
       - type: io.neonbee.endpoint.metrics.MetricsEndpoint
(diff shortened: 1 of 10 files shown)


src/main/java/io/neonbee/endpoint/odatav4/ODataV4Endpoint.java
```diff
@@ -243,4 +280,195 @@ public Router createEndpointRouter(Vertx vertx, String basePath, JsonObject conf
...
+            // parse out the resource path and entity name
+            resourcePath = requestPath.substring(routeMountPoint.length() + routePath.length() - 1);
+            entityName = emptyToNull(resourcePath.split("\\W", 2)[0]); // assume the first non-word char separates it
+
+            // if an entity name is provided, concatenate the full qualified name
+            fullQualifiedName = entityName != null ? schemaNamespace + '.' + entityName : null;
+
+            // construct the full request path and URI, the query is provided by the request
+            requestUri = hostUri + (this.requestPath = basePath + schemaNamespace + resourcePath)
+                    + (!(requestQuery = nullToEmpty(request.query())).isEmpty() ? "?" + requestQuery : EMPTY);
```
pk-work, line 463: this can be written less short which would increase readability.

src/main/java/io/neonbee/endpoint/odatav4/ODataV4Endpoint.java
```diff
@@ -243,4 +280,195 @@ public Router createEndpointRouter(Vertx vertx, String basePath, JsonObject conf
...
+            Route route = routingContext.currentRoute();
+            // note that getPath() returns *only* the path prefix, so essentially the base path and converted schema
+            // namespace with leading and tailing slashes w/o the tailing *, which is handled and stripped by the router
+            String routeMountPoint = routingContext.mountPoint();
+            String routePath = // routePath w/ exactly one tailing slash
+                    Optional.ofNullable(route).map(Route::getPath).orElse(EMPTY).replaceAll("/+$", EMPTY) + "/";
+
+            HttpServerRequest request = routingContext.request();
+            String requestPath = request.path();
+            if (!requestPath.contains(routePath)) {
```
pk-work, line 444: This is something I have to think about

resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
```diff
@@ -46,6 +46,10 @@ config:
         authenticationChain: ~
         # namespace and service name URI mapping (strict, or loose based on CDS)
         uriConversion: strict
+        # a block / allow list of verticles to expose via this endpoint (defaults to empty / all entities exposed)
```
pk-work, line 49: It should be mentioned that the value of block and allow attribute should be an array with regexp entries.

src/main/java/io/neonbee/endpoint/odatav4/ODataV4Endpoint.java
```diff
@@ -220,19 +238,38 @@ public Router createEndpointRouter(Vertx vertx, String basePath, JsonObject conf
...
+                                    // normalize the URI first
+                                    NormalizedUri normalizedUri = normalizeUri(routingContext, schemaNamespace);
+                                    if (LOGGER.isDebugEnabled()) {
+                                        LOGGER.correlateWith(routingContext).debug("Normalized OData V4 URI {}",
+                                                normalizedUri);
+                                    }
+
+                                    // if a entity is specified check it against the block list
+                                    // TODO: maybe also navigation properties have to be taken into account here?
+                                    if (normalizedUri.fullQualifiedName != null
```
pk-work, line 261: I'm not sure how expensive this regexp check is, but it will be performed for every request. The allow/block list cannot be changed after startup, maybe we should store the regexp results in a Map.

s4heid:
@pk-work @kristian fyi: ci is up and running again.

kristian:
I have checked on the test failures:

- testCDSUriConversion fails for me because `requestMetadata("test-service3")` should be changed to `requestMetadata("")`
- the failing testProhibitRequestsToUnderScoreDVsByDefault is a GOOD CATCH! In `RawEndpoint` the `HIDDEN_VERTICLES_PATTERN.matcher(qualifiedName).matches()` should be chaned to `HIDDEN_VERTICLES_PATTERN.matcher(qualifiedName).find()`. `matches()` always checks if the FULL STRING matches, which is not the case here...
- testMapToODataRequest fails due to a corner-case I did not think of in NormalizedUri, could you please change `resourcePath = requestPath.substring(routeMountPoint.length() + routePath.length() - 1);` to: `resourcePath = requestPath.substring((routeMountPoint + routePath).replaceAll("/{2,}", "/").length() - 1);`

This has fixed all issues for me. Thanks!

PR #50, opened 2021-05-21
feat: add support for custom error handlers

s4heid:
- rename ErrorHandler to DefaultErrorHandler
- remove create() pattern from the handler, because the
DefaultErrorHandler is a class and not an interface like the
ErrorHandler in Vert.x
- As in Vert.x an ErrorHandler interface already exists, reuse this
interface when implementing our own error handler.
- add errorHandler property to the ServerVerticleConfig

### DIFF
--- a/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
+++ b/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
@@ -34,6 +34,9 @@ config:
     # the maximum initial line length of the HTTP header (e.g. "GET / HTTP/1.0"), defaults to 4096 bytes
     maxInitialLineLength: 4096
 
+    # configure the error handler. If not specified, defaults to io.neonbee.internal.handler.ErrorHandler
+    errorHandler: io.neonbee.internal.handler.DefaultErrorHandler
+
     # specific endpoint configuration, defaults to the object seen below
     endpoints:
       # provides a OData V4 compliant endpoint, for accessing entity verticle data
--- a/src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
+++ b/src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
@@ -7,10 +7,10 @@ import static io.vertx.core.http.HttpHeaders.CONTENT_TYPE;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import io.neonbee.data.DataException;
 import io.neonbee.logging.LoggingFacade;
-import io.vertx.core.Handler;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.MIMEHeader;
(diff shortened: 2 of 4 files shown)


resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
```diff
@@ -34,6 +34,9 @@ config:
     # the maximum initial line length of the HTTP header (e.g. "GET / HTTP/1.0"), defaults to 4096 bytes
     maxInitialLineLength: 4096
 
+    # configure the error handler. If not specified, defaults to io.neonbee.internal.handler.ErrorHandler
+    errorHandler: io.neonbee.internal.handler.ErrorHandler
```
pk-work, line 38: DefaultErrorHandler?

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -77,6 +76,71 @@ public void handle(RoutingContext routingContext) {
...
+     * @param errorCode      the error code of the response
+     * @param errorMessage   the error message of the response
+     * @return A String with the error response in JSON format.
+     */
+    protected String createJsonResponse(RoutingContext routingContext, int errorCode, String errorMessage) {
+        JsonObject jsonResponse = new JsonObject().put("code", errorCode).put("message", errorMessage);
+        String correlationId = getCorrelationId(routingContext);
+        if (correlationId != null) {
+            jsonResponse.put("correlationId", correlationId);
+        }
```
pk-work, line 110: What do you think about:

Optional.ofNullable(getCorrelationId(routingContext)).ifPresent(id -> jsonResponse.put("correlationId", id ))

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -77,6 +76,71 @@ public void handle(RoutingContext routingContext) {
...
+    /**
+     * Creates a String with the application/json error response.
+     *
+     * @param routingContext the routing context
+     * @param errorCode      the error code of the response
+     * @param errorMessage   the error message of the response
+     * @return A String with the error response in JSON format.
+     */
+    protected String createJsonResponse(RoutingContext routingContext, int errorCode, String errorMessage) {
+        JsonObject jsonResponse = new JsonObject().put("code", errorCode).put("message", errorMessage);
```
pk-work, line 106: jsonResponse -> error? or errorObject, because the response is the whole thing you return when calling encode in line 111

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -77,6 +76,71 @@ public void handle(RoutingContext routingContext) {
...
+
+    /**
+     * Creates a String with the application/json error response.
+     *
+     * @param routingContext the routing context
+     * @param errorCode      the error code of the response
+     * @param errorMessage   the error message of the response
+     * @return A String with the error response in JSON format.
+     */
+    protected String createJsonResponse(RoutingContext routingContext, int errorCode, String errorMessage) {
```
pk-work, line 105: this method should return json. Then you can also change, there is a method for sending JSON which also directly sets the content type, if I remember correctly.

if (mime.startsWith("application/json")) {
            response.putHeader(CONTENT_TYPE, "application/json");
            response.end(createJsonResponse(routingContext, errorCode, errorMessage));

pk-work (changes requested):

pk-work (approved):

PR #60, opened 2021-07-16
refactor: move error handler configuration to ServerConfig class

kristian:
Before this change the JsonObject passed to the ServerVerticle was used to read the configuration for the custom error template to use. With this change the JsonObject was replaced by the ServerConfig object. In addition to a custom errorHandlerClassName a new errorHandlerTemplate property was added, which can be used to just specify a different (HTML) template to use in the DefaultErrorHandler.

### DIFF
--- a/src/main/java/io/neonbee/config/ServerConfig.java
+++ b/src/main/java/io/neonbee/config/ServerConfig.java
@@ -19,6 +19,7 @@ import com.google.common.collect.ImmutableBiMap;
 import io.neonbee.endpoint.metrics.MetricsEndpoint;
 import io.neonbee.endpoint.odatav4.ODataV4Endpoint;
 import io.neonbee.endpoint.raw.RawEndpoint;
+import io.neonbee.internal.handler.DefaultErrorHandler;
 import io.neonbee.internal.json.ImmutableJsonObject;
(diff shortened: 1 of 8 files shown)


src/main/java/io/neonbee/config/ServerConfig.java
```diff
@@ -398,6 +404,60 @@ public ServerConfig setAuthChainConfig(List<AuthHandlerConfig> authChainConfig)
...
+     *
+     * @return the file name of a error handler template, or null. in latter case the error handler has the choice of
+     *         the template to use
+     */
+    public String getErrorHandlerTemplate() {
+        return errorHandlerTemplate;
+    }
+
+    /**
+     * Sets the path to a error handler template to use either in the {@link DefaultErrorHandler} or in any custom error
```
pk-work, line 448:
```suggestion
     * Sets the path to an error handler template to use either in the {@link DefaultErrorHandler} or in any custom error
```

pk-work (changes requested):
Only typos

kristian:
> Only typos

Thanks for the review Pascal, I did fix the typos

pk-work (approved):

PR #64, opened 2021-08-11
feat: add JobSchedule validation

pk-work:
A periodic JobSchedule with an interval of 0 could result into an
endless loop when JobVerticle tries to calculate the next execution.

### DIFF
--- a/src/main/java/io/neonbee/job/JobVerticle.java
+++ b/src/main/java/io/neonbee/job/JobVerticle.java
@@ -25,6 +25,8 @@ public abstract class JobVerticle extends AbstractVerticle {
 
     private static final long MINIMUM_DELAY = 100L;
 
+    private static final int SCHEDULE_TEST_EXECUTIONS = 10;
+
     private JobSchedule schedule;
 
     private Instant lastExecution;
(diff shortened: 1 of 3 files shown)


igarab (dismissed):

src/main/java/io/neonbee/job/JobVerticle.java
```diff
@@ -74,12 +74,20 @@ public String getName() {
     @Override
     public void start() {
         if (!NeonBee.get(getVertx()).getOptions().shouldDisableJobScheduling()) {
+            if (schedule.isPeriodic() && isTemporalAdjusterInvalid()) {
```
kristian, line 77: isScheduleValid()

or

validateSchedule()

so we can add further checks to it later on?

src/main/java/io/neonbee/job/JobVerticle.java
```diff
@@ -74,12 +74,20 @@ public String getName() {
...
+                throw new IllegalStateException("JobVerticle was configured with an invalid JobSchedule");
```
kristian, line 78: Very technical error message with class names, is ir better to just call it an "invalid schedule" and explain why it is invalid

src/main/java/io/neonbee/job/JobVerticle.java
```diff
@@ -74,12 +74,20 @@ public String getName() {
...
+            }
             scheduleJob();
         } else {
             finalizeJob();
         }
     }
 
+    private boolean isTemporalAdjusterInvalid() {
+        Instant testInstant = now();
+        return testInstant.with(schedule).compareTo(testInstant) <= 0;
```
kristian, line 88: Mabe we should even call it multiple times, to see if it adjuste forward every time, just to be sure.

src/main/java/io/neonbee/job/JobVerticle.java
pk-work: done

src/main/java/io/neonbee/job/JobVerticle.java
pk-work: done

src/main/java/io/neonbee/job/JobVerticle.java
pk-work: done

pk-work:
voter is not working again .. but change runs through locally and @kristian and @SAPDaniloWork has reviewed the change, so I will merge it.

PR #65, opened 2021-08-12
chore: don't start gradle as a daemon

pk-work:


kristian:
Minimal changes like this don't require a review, esp. because we are a very small committers group atm. Let's discuss this innour weekly comitters meeting.

kristian (approved):

PR #66, opened 2021-08-23
fix: rephrase error handler properties in config

s4heid:
With this fix, the properties `errorHandler` and `errorTemplate`, which
can be configured in the ServerVerticle config, will be working again. A
regression in the ServerConfig caused that these properties are ignored.


kristian (approved):

kristian:
LGTM. Thanks @s4heid!

pk-work (approved):

PR #67, opened 2021-09-03
feat: add session id to DataContext

pk-work:


s4heid (approved):

s4heid:
it's friday -> voter tired. voter needs sleep... merging as tests were green locally.

kristian:
Good change, thanks @pk-work. Thanks for the review @s4heid!

PR #80, opened 2021-11-17
feat: offer a new ErrorHandler which can be initialized asynchronously

pk-work:
A class that implements the new ErrorHandler interface MUST have a
default constructor. Otherwise NeonBee can't instantiate the class
during bootstrap phase.

The purpose of the initialize method is to allow the execution of
blocking code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -191,7 +191,8 @@ clean {
 // ############ Test Phase
 
 test {
-    if (!org.gradle.nativeplatform.platform.internal.DefaultNativePlatform.getCurrentOperatingSystem().isMacOsX()) {
+    // deactivate parallel execution of tests temporary until parallel execution works again
+    if (false && !org.gradle.nativeplatform.platform.internal.DefaultNativePlatform.getCurrentOperatingSystem().isMacOsX()) {
         systemProperties = [
             'junit.jupiter.execution.parallel.enabled' : 'true',
             'junit.jupiter.execution.parallel.mode.default' : 'concurrent',
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.annotations.VisibleForTesting;
 
 import io.neonbee.config.NeonBeeConfig;
+import io.neonbee.config.ServerConfig;
 import io.neonbee.data.DataQuery;
 import io.neonbee.entity.EntityWrapper;
 import io.neonbee.hook.HookRegistry;
@@ -590,6 +591,15 @@ public class NeonBee {
      */
     public void unregisterLocalConsumer(String verticleAddress) {
         localConsumers.remove(verticleAddress);
+    }
+
+    /**
+     * Returns the ServerConfig if NeonBee is started with WEB profile.
+     *
+     * @return the ServerConfig or null if ServerVerticle is not started.
+     */
+    public ServerConfig getServerConfig() {
+        return new ServerConfig((JsonObject) getLocalMap().get(ServerVerticle.SERVER_CONFIG_KEY));
     }
 
     /**
--- a/src/main/java/io/neonbee/handler/ErrorHandler.java
+++ b/src/main/java/io/neonbee/handler/ErrorHandler.java
@@ -1,0 +1,21 @@
+package io.neonbee.handler;
+
+import io.neonbee.NeonBee;
+import io.vertx.core.Future;
+
+/**
+ * The class which implements this interface MUST have a default constructor. Otherwise NeonBee can't instantiate the
+ * class during bootstrap phase.
+ */
+@SuppressWarnings("NM_SAME_SIMPLE_NAME_AS_INTERFACE")
+public interface ErrorHandler extends io.vertx.ext.web.handler.ErrorHandler {
+
+    /**
+     * The purpose of this method is to offer the possibility to also executed blocking code during the initialization
+     * of the ErrorHandler. This is beneficial in case that the error template needs to be read from a database or file.
+     *
+     * @param neonBee The related NeonBee instance
+     * @return a Future with a reference of this ErrorHandler
+     */
+    Future<ErrorHandler> initialize(NeonBee neonBee);
+}
--- a/src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
+++ b/src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
@@ -1,17 +1,20 @@
 package io.neonbee.internal.handler;
 
 import static io.neonbee.internal.handler.CorrelationIdHandler.getCorrelationId;
-import static io.neonbee.internal.helper.BufferHelper.readResourceToBuffer;
 import static io.netty.handler.codec.http.HttpResponseStatus.INTERNAL_SERVER_ERROR;
 import static io.vertx.core.http.HttpHeaders.CONTENT_TYPE;
 
-import java.io.IOException;
+import java.nio.file.Path;
 import java.util.List;
-import java.util.Objects;
 import java.util.Optional;
 
+import io.neonbee.NeonBee;
 import io.neonbee.data.DataException;
+import io.neonbee.handler.ErrorHandler;
+import io.neonbee.internal.helper.FileSystemHelper;
 import io.neonbee.logging.LoggingFacade;
+import io.vertx.core.Future;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.MIMEHeader;
@@ -20,11 +23,11 @@ import io.vertx.ext.web.RoutingContext;
 /**
  * Similar to the io.vertx.ext.web.handler.impl.ErrorHandlerImpl, w/ minor adoptions for error text and template.
  */
-public class DefaultErrorHandler implements io.vertx.ext.web.handler.ErrorHandler {
+public class DefaultErrorHandler implements ErrorHandler {
     /**
      * The default template to use for rendering.
      */
-    public static final String DEFAULT_ERROR_HANDLER_TEMPLATE = "META-INF/neonbee/error-template.html";
+    public static final Path DEFAULT_ERROR_HANDLER_TEMPLATE = Path.of("META-INF/neonbee/error-template.html");
 
     private static final LoggingFacade LOGGER = LoggingFacade.create();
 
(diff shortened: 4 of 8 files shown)


src/main/java/io/neonbee/handler/ErrorHandler.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * The purpose of this method is to offer the possibility to also executed blocking code during the initialization
+     * of the ErrorHandler. This is beneficial in case that the error template needs to be read from a database or file.
+     *
+     * @param neonbee           The related NeonBee instance
+     * @param errorTemplateName The name of the error template
+     * @return a Future with a reference of this ErrorHandler
+     */
+    Future<ErrorHandler> initialize(NeonBee neonbee, String errorTemplateName);
```
kristian, line 20: Better to use Future<Void> here.
Also should we rather use Vertx vertx, instead of NeonBee? And is it NeonBee neonBee, oder neonBee?
Also one additional thing came to my mind:
The errorTemplateName could stay part of the constructor, while the initialize then is reduced to simply:
Future<Void> initialize(Vertx vertx) or Future<Void> initialize(NeonBee neonBee)

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -35,31 +38,35 @@
...
      * @param errorTemplateName resource path to the template.
-     * @throws IOException in case the template is not found or cannot be read from the current class loader.
+     *
+     * @return A Future with the initialized ErrorHandler
      */
-    public DefaultErrorHandler(String errorTemplateName) throws IOException {
-        this.errorTemplate = readErrorTemplate(errorTemplateName);
+    @Override
+    public Future<ErrorHandler> initialize(NeonBee neonbee, String errorTemplateName) {
+        return AsyncHelper.<String>executeBlocking(neonbee.getVertx(), p -> {
```
kristian, line 53: No blocking read is required any longer, please use a non blocking read here

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -76,57 +76,15 @@ public void handle(RoutingContext routingContext) {
...
-                                    LOGGER.error("HTTP server could not be started", cause);
-                                }).<Void>mapEmpty().onComplete(startPromise);
-                    });
-        } catch (Exception e) {
-            LOGGER.error("Server could not be started", e);
-            startPromise.fail(e);
-        }
+        createRouter(config).compose(router -> {
+            Optional<AuthenticationHandler> ach = createAuthChainHandler(config.getAuthChainConfig());
+            return mountEndpoints(router, config.getEndpointConfigs(), ach, new HooksHandler()).map(v -> {
```
kristian, line 81: Instead of using map(v->) use onSuccess

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -137,27 +95,78 @@ public void stop(Promise<Void> stopPromise) throws Exception {
...
+                    rootRoute.handler(new CacheControlHandler());
+                    rootRoute.handler(new InstanceInfoHandler());
+
+                    createSessionStore(vertx, config.getSessionHandling()).map(SessionHandler::create)
+                            .ifPresent(sessionHandler -> rootRoute
+                                    .handler(sessionHandler.setSessionCookieName(config.getSessionCookieName())));
+
+                    return succeededFuture(router);
+                }).onFailure(e -> {
+                    LOGGER.error("Server could not be started", e);
```
kristian, line 123: This is a wrong error message, the thing that fail here, is the error handler which could not be created and it should be checked with .onFailure only with the error handler.

kristian (changes requested):

src/main/java/io/neonbee/handler/ErrorHandler.java
pk-work: If I want to write y custom handler that wants to benefit from a errorTemplateName, how du I know that this is possible? Of course I can read the JavaDoc and maybe don't miss the point that I'm able to do this. But if it is part of the signature I have to implement, it is more visible to me.

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
pk-work: We don't have an async BufferHelper.readResourceToBuffer. Should I create one? But then I would wrap if with AsynHelper again because there is no async offering of Classloader.getResourceAsStream

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
pk-work: I used map to ensure that the handler is added before createHttpServer is called.  Would it be a problem in case that the handler is added afterwards?

src/main/java/io/neonbee/handler/ErrorHandler.java
pk-work: I passed Vert.x in the first version, but then I thought: "what if I need a config from the config directory to initialize the handler?" Of course I can do NeonBee.get(vertx), but I can also do neonBee.gertVertx() :)

I don't have strong feelings here, I will do what you prefer :)

pk-work: We agreed on passing only NeonBee and add a getServerConfig() method to NeonBee. errorTemplateName can then be retrieved via the ServerConfig.

ServerConfig will be stored in the NeonBee context

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
pk-work: I've tested it, it doesn't matter. Will use onSuccess

build.gradle
```diff
@@ -191,7 +191,8 @@ clean {
 // ############ Test Phase
 
 test {
-    if (!org.gradle.nativeplatform.platform.internal.DefaultNativePlatform.getCurrentOperatingSystem().isMacOsX()) {
+    // deactivate parallel execution of tests temporary until parallel execution works again
+    if (false && !org.gradle.nativeplatform.platform.internal.DefaultNativePlatform.getCurrentOperatingSystem().isMacOsX()) {
```
kristian, line 195: I think you forgot to remove false, or?

src/main/java/io/neonbee/NeonBee.java
```diff
@@ -592,6 +593,15 @@ public void unregisterLocalConsumer(String verticleAddress) {
         localConsumers.remove(verticleAddress);
     }
 
+    /**
+     * Returns the ServerConfig if NeonBee is started with WEB profile.
+     *
+     * @return the ServerConfig or null if ServerVerticle is not started.
+     */
+    public ServerConfig getServerConfig() {
+        return new ServerConfig((JsonObject) getLocalMap().get(ServerVerticle.SERVER_CONFIG_KEY));
```
kristian, line 602: Instead of storing the JSON object, how about storing the ServerConfig directly, that way it will not need to get parsed everytime again.

src/main/java/io/neonbee/handler/ErrorHandler.java
```diff
@@ -0,0 +1,21 @@
...
+import io.neonbee.NeonBee;
+import io.vertx.core.Future;
+
+/**
+ * The class which implements this interface MUST have a default constructor. Otherwise NeonBee can't instantiate the
+ * class during bootstrap phase.
+ */
+@SuppressWarnings("NM_SAME_SIMPLE_NAME_AS_INTERFACE")
+public interface ErrorHandler extends io.vertx.ext.web.handler.ErrorHandler {
+
```
kristian, line 12: Can we add a rule to not have this "initial blank line" in all classes? It drives me nuts! :D Thanks.

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -35,31 +38,27 @@
...
+     * @param neonbee the related NeonBee instance
      *
-     * @param errorTemplateName resource path to the template.
-     * @throws IOException in case the template is not found or cannot be read from the current class loader.
+     * @return A Future with the initialized ErrorHandler
      */
-    public DefaultErrorHandler(String errorTemplateName) throws IOException {
-        this.errorTemplate = readErrorTemplate(errorTemplateName);
+    @Override
+    public Future<ErrorHandler> initialize(NeonBee neonbee) {
```
kristian, line 51: neonBee

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -35,31 +38,27 @@
...
+        return readErrorTemplate(neonbee).map(errorTemplate -> {
```
kristian, line 52: If you return this; in .map you can also use onSuccess with the same result

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
```diff
@@ -35,31 +38,27 @@
...
+            this.errorTemplate = errorTemplate;
+            return this;
+        });
     }
 
-    private String readErrorTemplate(String errorTemplateName) throws IOException {
-        return readResourceToBuffer(Objects.requireNonNull(errorTemplateName)).toString();
+    private Future<String> readErrorTemplate(NeonBee neonBee) {
+        String errorTemplatePath = neonBee.getServerConfig().getErrorHandlerTemplate();
+        Path p = errorTemplatePath == null ? DEFAULT_ERROR_HANDLER_TEMPLATE : Path.of(errorTemplatePath);
```
kristian, line 60: Better use a Optional.ofNullable here and make it one statement?

Optional.ofNullable(neonBee.getServerConfig().getErrorHandlerTemplate()).map(Path::of).orElse(DEFAULT_ERROR_TEMPLATE)

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -98,66 +123,56 @@ public void start(Promise<Void> startPromise) {
...
     // reason for PMD.EmptyCatchBlock empty cache block is "path-through" to fallback implementation and for
     // PMD.SignatureDeclareThrowsException it does not matter, if this private method does not expose any concrete
     // exceptions as the ServerVerticle start method anyways catches all exceptions in a generic try-catch block
     @SuppressWarnings({ "PMD.EmptyCatchBlock", "PMD.SignatureDeclareThrowsException" })
-    static ErrorHandler createErrorHandler(String className, String template) throws Exception {
-        Class<?> classObject = Class.forName(Optional.ofNullable(className).filter(Predicate.not(String::isBlank))
-                .orElse(DEFAULT_ERROR_HANDLER_CLASS_NAME));
-        if (template != null) {
+    static Future<ErrorHandler> createErrorHandler(String className, Vertx vertx) {
+        return AsyncHelper.executeBlocking(vertx, p -> {
```
kristian, line 138: Is a executeBlcoking still required here?

Calling the constructor should be pretty quick and the eh.initialize method is now asynchronous, thus I would think we can remove the blocking call to not cause worker threads to be utilized even more during boot.

src/main/java/io/neonbee/internal/verticle/ServerVerticle.java
```diff
@@ -74,18 +80,37 @@ public void handle(RoutingContext routingContext) {
 
     @Override
     public void start(Promise<Void> startPromise) {
+        NeonBee.get(vertx).getLocalMap().put(SERVER_CONFIG_KEY, config());
```
kristian, line 83: Let's move this down 2 rows and then put the parsed "config" object in here 👍

kristian (changes requested):

src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java
halber: That is not what Optional is intended for. Using optional here would allocate an object where no object allocation is necessary and to quote Stuart Marks: „It’s generally a bad idea to create an Optional for the specific purpose of chaining methods from it to get a value.“

kristian: I can see the point of Stuart here. Interesting consideration! Thanks for bringing this up. I think we should add that as a rule to our coding guidelines. I think Optional is used quite broadly in many methods of us.

s4heid (approved):

PR #86, opened 2021-12-13
feat: move MetricOptions to NeonBeeOptions

halber:
By moving the MetricOptions to the NeonBeeOptions, we are able to change the MetricOptions in a LauncherPreProcessor.

### DIFF
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -73,8 +73,6 @@ import io.vertx.core.impl.ConcurrentHashSet;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.shareddata.AsyncMap;
 import io.vertx.core.shareddata.LocalMap;
-import io.vertx.micrometer.MicrometerMetricsOptions;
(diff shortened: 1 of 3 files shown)


sap-C5335577 (approved):

kristian (approved):
lgtm

pk-work (approved):

kristian:
Was there a reason why we add this to Options and not to Config?

So basis question: Should this be configurable at runtime, or with a config file?

halber:
> Was there a reason why we add this to Options and not to Config?
> 
> So basis question: Should this be configurable at runtime, or with a config file?

The Problem ist, that we need this configuration to create a new Vertx instance. Loading the NeonBeeConfig happens after the instance has been created.

kristian:
> > Was there a reason why we add this to Options and not to Config?
> > So basis question: Should this be configurable at runtime, or with a config file?
> 
> The Problem ist, that we need this configuration to create a new Vertx instance. Loading the NeonBeeConfig happens after the instance has been created.

Good point. So my "lgtm" stays. I think it is fine to be merged in. Good addition. Thanks!

PR #94, opened 2021-12-22
feat: add custom micrometer registries via NeonBeeConfig

s4heid:
With this change the config parameter `micrometerRegistries` is
introduced which allows the user to specify a list of (full qualified) class
names, which must implement the `load()` method of the functional interface
`MicrometerRegistryLoader`. The load method must return a
`MeterRegistry`, which will be added to the micrometer registries by
NeonBee. The `PrometheusMeterRegistry` will be the default
registry, which is only available when the `MetricsEndpoint` is loaded.

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,70 @@
+# Metrics Concept
+
+## Content
+
+- [Metrics](#Metrics)
+  - [configure additional metrics registries](#configure-additional-metrics-registries)
+    - [when launching NeonBee](#when-launching-NeonBee)
+    - [during runtime](#during-runtime)
+  - [add metrics to your code](#add-metrics-to-your-code)
+
+
+## Metrics
+In NeonBee, Micrometer is used to provide reporting to various backends. Micrometer provides a facade for the most
+common monitoring systems.
+Micrometer `MeterRegistry` objects can be registered at runtime. For this purpose, a `CompositeMeterRegistry` is added in
+the `VertxOptions`. Additional `MeterRegistry` can be added to this `CompositeMeterRegistry` at runtime.
+
+The `MetricsEndpoint`, which by default provides the Prometheus metrics under the /metrics path, registers the
+`PrometheusMeterRegistry` when the `MetricsEndpoint` router is created.
+
+## configure additional metrics registries
+### when launching NeonBee
+To register own Micrometer MeterRegistry interface `MicrometerRegistryLoader` must be implemented
+and the implementing class must be specified in the configuration `io.neonbee.NeonBee.yaml` in the `micrometerRegistries`
+array.
+
+Example MicrometerRegistryLoader implementation:
+```java
+package io.neonbee.config.examples;
+
+import io.micrometer.core.instrument.logging.LoggingMeterRegistry;
+import io.neonbee.config.metrics.MicrometerRegistryLoader;
+
+public static class LoggingMeterMicrometerRegistryLoader implements MicrometerRegistryLoader {
+    @Override
+    public MeterRegistry load(JsonObject config) {
+        return new LoggingMeterRegistry();
+    }
+}
+```
+
+Example io.neonbee.NeonBee.yaml configuration:
+```yaml
+---
+  // Omitted other configuration values
+  micrometerRegistries:
+  - className: io.neonbee.config.examples.LoggingMeterMicrometerRegistryLoader
+    config:
+      key: value
+```
+
+### during runtime
+If you want to add a MeterRegistry during runtime, you can do it using the `NeonBeeConfig#getCompositeMeterRegistry()`
+method.
+
+```java
+MeterRegistry yourRegistry; // Your registry to be added.
+CompositeMeterRegistry compositeMeterRegistry = NeonBee.get(vertx).getConfig().getCompositeMeterRegistry();
+compositeMeterRegistry.add(yourRegistry);
+```
+## add metrics to your code
+
+To provide metrics in your code, here is an example of a counter:
+```java
+MeterRegistry registry = BackendRegistries.getDefaultNow();
+Counter counter = registry.counter("TestEndpointCounter", "TestTag1", "TestValue");
+counter.increment();
+count = counter.count();
+```
+For more information, see [user defined metrics](https://vertx.io/docs/vertx-micrometer-metrics/java/#_user_defined_metrics)
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -10,6 +10,7 @@ import static io.vertx.core.Future.succeededFuture;
 import static java.lang.System.setProperty;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -22,7 +23,6 @@ import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import io.neonbee.config.NeonBeeConfig;
 import io.neonbee.config.ServerConfig;
 import io.neonbee.data.DataQuery;
@@ -73,6 +74,7 @@ import io.vertx.core.impl.ConcurrentHashSet;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.shareddata.AsyncMap;
 import io.vertx.core.shareddata.LocalMap;
+import io.vertx.micrometer.MicrometerMetricsOptions;
 import io.vertx.spi.cluster.hazelcast.HazelcastClusterManager;
 
 public class NeonBee {
@@ -153,12 +155,14 @@ public class NeonBee {
 
     private final Set<String> localConsumers = new ConcurrentHashSet<>();
 
+    private final CompositeMeterRegistry compositeMeterRegistry;
+
     /**
      * Convenience method for returning the current NeonBee instance.
      * <p>
      * Important: Will only return a value in case a Vert.x context is available, otherwise returns null. Attention:
      * This method is NOT signature compliant to {@link Vertx#vertx()}! It will NOT create a new NeonBee instance,
-     * please use {@link NeonBee#create(NeonBeeOptions)} or {@link NeonBee#create(Supplier, NeonBeeOptions)} instead.
+     * please use {@link NeonBee#create(NeonBeeOptions)} or {@link NeonBee#create(Function, NeonBeeOptions)} instead.
      *
      * @return A NeonBee instance or null
      */
@@ -195,12 +199,12 @@ public class NeonBee {
      * @return the future to a new NeonBee instance initialized with default options and a new Vert.x instance
      */
     public static Future<NeonBee> create(NeonBeeOptions options) {
-        return create((OwnVertxSupplier) () -> newVertx(options), options);
+        return create((OwnVertxFactory) (vertxOptions) -> newVertx(vertxOptions, options), options);
     }
 
     @VisibleForTesting
     @SuppressWarnings({ "PMD.EmptyCatchBlock", "PMD.AvoidCatchingThrowable" })
-    static Future<NeonBee> create(Supplier<Future<Vertx>> vertxFutureSupplier, NeonBeeOptions options) {
+    static Future<NeonBee> create(Function<VertxOptions, Future<Vertx>> vertxFactory, NeonBeeOptions options) {
         try {
             // create the NeonBee working and logging directory (as the only mandatory directory for NeonBee)
             Files.createDirectories(options.getLogDirectory());
@@ -217,12 +221,19 @@ public class NeonBee {
         InternalLoggerFactory.setDefaultFactory(Slf4JLoggerFactory.INSTANCE);
         logger = LoggerFactory.getLogger(NeonBee.class);
 
+        VertxOptions vertxOptions = new VertxOptions().setEventLoopPoolSize(options.getEventLoopPoolSize())
+                .setWorkerPoolSize(options.getWorkerPoolSize());
+
+        CompositeMeterRegistry compositeMeterRegistry = new CompositeMeterRegistry();
+        vertxOptions.setMetricsOptions(
+                new MicrometerMetricsOptions().setMicrometerRegistry(compositeMeterRegistry).setEnabled(true));
+
         // create a Vert.x instance (clustered or unclustered)
-        return vertxFutureSupplier.get().compose(vertx -> {
+        return vertxFactory.apply(vertxOptions).compose(vertx -> {
             // at this point at any failure that occurs, it is in our responsibility to properly close down the created
             // Vert.x instance again. we have to be vigilant the fact that a runtime exception could happen anytime!
             Function<Throwable, Future<Void>> closeVertx = throwable -> {
-                if (!(vertxFutureSupplier instanceof OwnVertxSupplier)) {
+                if (!(vertxFactory instanceof OwnVertxFactory)) {
                     // the Vert.x instance is *not* owned by us, thus don't close it either
                     logger.error("Failure during bootstrap phase.", throwable); // NOPMD slf4j
                     return failedFuture(throwable);
@@ -238,10 +249,18 @@ public class NeonBee {
 
             try {
                 // create a NeonBee instance, hook registry and close handler
-                NeonBee neonBee = new NeonBee(vertx, options);
+                NeonBee neonBee = new NeonBee(vertx, options, compositeMeterRegistry);
 
                 // load the configuration and boot it up, on failure close Vert.x
-                return neonBee.loadConfig().compose(config -> neonBee.boot()).recover(closeVertx).map(neonBee);
+                return neonBee.loadConfig().compose(config -> {
+                    try {
+                        config.createMicrometerRegistries().forEach(neonBee.compositeMeterRegistry::add);
+                        return succeededFuture(config);
+                    } catch (ClassNotFoundException | NoSuchMethodException | InvocationTargetException
+                            | InstantiationException | IllegalAccessException e) {
+                        return failedFuture(e);
+                    }
+                }).compose(config -> neonBee.boot()).recover(closeVertx).map(neonBee);
             } catch (Throwable t) {
                 // on any exception (e.g. during the initialization of a NeonBee object) don't forget to close Vert.x!
                 return closeVertx.apply(t).mapEmpty();
@@ -250,10 +269,7 @@ public class NeonBee {
     }
 
     @VisibleForTesting
-    static Future<Vertx> newVertx(NeonBeeOptions options) {
-        VertxOptions vertxOptions = new VertxOptions().setEventLoopPoolSize(options.getEventLoopPoolSize())
-                .setWorkerPoolSize(options.getWorkerPoolSize()).setMetricsOptions(options.getMetricsOptions());
-
+    static Future<Vertx> newVertx(VertxOptions vertxOptions, NeonBeeOptions options) {
         if (!options.isClustered()) {
             return succeededFuture(Vertx.vertx(vertxOptions));
         } else {
@@ -470,9 +486,10 @@ public class NeonBee {
     }
 
     @VisibleForTesting
-    NeonBee(Vertx vertx, NeonBeeOptions options) {
+    NeonBee(Vertx vertx, NeonBeeOptions options, CompositeMeterRegistry compositeMeterRegistry) {
         this.vertx = vertx;
         this.options = options;
+        this.compositeMeterRegistry = compositeMeterRegistry;
 
         // to be able to retrieve the NeonBee instance from any point you have a Vert.x instance add it to a global map
         NEONBEE_INSTANCES.put(vertx, this);
@@ -480,7 +497,8 @@ public class NeonBee {
         registerCloseHandler(vertx);
     }
 
-    private Future<NeonBeeConfig> loadConfig() {
+    @VisibleForTesting
+    Future<NeonBeeConfig> loadConfig() {
         return NeonBeeConfig.load(vertx).onSuccess(config -> this.config = config);
     }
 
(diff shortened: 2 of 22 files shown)


src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -205,4 +266,27 @@ public NeonBeeConfig setTimeZone(String timeZone) {
         this.timeZone = timeZone;
         return this;
     }
+
+    /**
+     * Adds the passed list of {@link String}s containing the name of classes which implement the
```
pk-work, line 271: FQN of classes

src/main/java/io/neonbee/internal/helper/FileSystemHelper.java
```diff
@@ -99,28 +102,55 @@ private FileSystemHelper() {}
...
+     * @return the YAML file converted to an {@link JsonObject}
+     * @throws IOException if an I/O error occurs reading from the file or a malformed or unmappable byte sequence is
+     *                     read
+     */
+    public static JsonObject readYAML(Path path) throws IOException {
+        return parseYAML(Files.readAllBytes(path));
+    }
+
     private static Future<JsonObject> parseYAML(Vertx vertx, Buffer buffer) {
         return vertx.executeBlocking(promise -> {
```
pk-work, line 141: Use executeBlocking from AsyncHelper?

src/test/java/io/neonbee/config/NeonBeeConfigTest.java
```diff
@@ -48,16 +64,65 @@ protected WorkingDirectoryBuilder provideWorkingDirectoryBuilder(TestInfo testIn
...
             testContext.completeNow();
         }));
     }
 
+    @Test
+    @Timeout(value = 2, timeUnit = TimeUnit.SECONDS)
+    @DisplayName("should load NeonBeeConfig correctly from working dir")
+    @SuppressWarnings("SuppressWarningsDeprecated")
+    void testLoadNeonBeeConfigBlocking() throws Exception {
+        NeonBeeConfig config = NeonBeeConfig.load(Path.of("src", "test", "resources", "io", "neonbee", "config"));
```
pk-work, line 118: TEST_RESOURCES.resolve("io/neonbee/config")

or

Path.of("src/test/resources/io/neonbee/config")

pk-work (changes requested):
This change is doing far more then only load the NeonBeeConfig earlier during the NeonBee creation process. It introduces a complete new mechanism to register metrics.

You should describe this in the commit message, or even add some docu how I can use this new mechanism. Whats the benefit of creating own registries? Can I add with this my own metrics? Are they automatically visible to the endpoint?

src/main/java/io/neonbee/internal/helper/FileSystemHelper.java
s4heid: The `AsyncHelper.executeBlocking(...)` takes a vertx instance which we cannot provide when we calling the method.

src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -49,16 +58,37 @@
...
 
+    /**
+     * Loads the NeonBee configuration (blocking) from the NeonBee config directory and creates an instance of
+     * {@link NeonBeeConfig}.
+     *
+     * @param path The path to the config directory of NeonBee
+     * @return A {@link NeonBeeConfig} configured by config file.
+     * @throws IOException if reading from the config file fails.
+     */
+    public static NeonBeeConfig load(Path path) throws IOException {
```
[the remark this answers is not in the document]
s4heid: I added the note to the javadoc.

src/test/java/io/neonbee/config/NeonBeeConfigTest.java
s4heid: makes sense to use the TEST_RESOURCES resolver

src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -75,6 +105,37 @@ public NeonBeeConfig(JsonObject json) {
...
     }
 
+    /**
+     * Load the default {@link PrometheusMeterRegistry} and try to load all {@link MicrometerRegistryLoader}s which are
+     * configured in the {@link NeonBeeConfig}.
+     *
+     * @param config the {@link NeonBeeConfig}
+     * @return MicrometerMetricsOptions which contains the loaded registries
+     */
+    public static MicrometerMetricsOptions loadMicrometerOptions(NeonBeeConfig config) {
```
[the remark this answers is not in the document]
s4heid: I extracted the mechanism of adding custom registries into its own method. I think this makes the code more readable. Also the method name is renamed from `loadMicrometerOptions()` to `createMicrometerOptions()`.

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
```diff
@@ -0,0 +1,56 @@
...
+        MeterRegistry meterRegistry;
+        if (registryName == null) {
+            meterRegistry = BackendRegistries.getDefaultNow();
+        } else {
+            meterRegistry = BackendRegistries.getNow(registryName);
+        }
+
+        if (meterRegistry instanceof CompositeMeterRegistry) {
+            CompositeMeterRegistry cmr = (CompositeMeterRegistry) meterRegistry;
+            cmr.getRegistries().stream().filter(registry -> registry instanceof PrometheusMeterRegistry).findAny()
```
pk-work, line 44: We should filter by ID otr name or something like that, to ensure that we ALWAYS get the default PrometheusMeterRegistry. Otherwise we don't know which PrometheusMeterRegistry we get in case that we also use one or multiple custom regestries of type PrometheusMeterRegistry

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
```diff
@@ -0,0 +1,56 @@
...
+                    .map(PrometheusMeterRegistry.class::cast)
+                    .ifPresentOrElse(pmr -> handleWithPrometheusMeterRegistry(rc, pmr), () -> LOGGER
```
pk-work, line 46: in case of error we need to fail the rc, otherwise request hangs forever.

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
halber: I can't filter the PrometheusMeterRegistry at this point and would therefore suggest that if you configure your own PrometheusMeterRegistry the default PrometheusMeterRegistry in the NeonBeeConfig is not added. If someone really has the case that he has two PrometheusMeterRegistry he would have to solve this himself by writing code.

kristian:
Hey everyone! Happy new year! 🥂🎆 

@halber I really like where this change is going and it introduces a lot of simplifications / nicer signatures that I like. Good work so far! I was about to do my initial review on the change, but I unfortunately have to open another round of discussion here for everyone:

@pk-work, @s4heid and @halber After reviewing the whole change, what I really do dislike is, that we previously made efforts of removing all blocking code. Now with this change we are re-introducing it across multiple classes like NeonBeeConfig, the ConfigHelper etc, alongside of all the new code that is needed to test the synchronous stuff... With that being said, wouldn't be better of doing it like it was suggested in vertx-config in the first place? See [1]. More specifically (to compare):

We could introduce a empty "load()" method in `NeonBeeConfig` that intern loads a NeonBeeConfig using a new Vert.x instance and closes it immediately, alongside:

```java
public static Future<NeonBeeConfig> load() {
   Vertx vertx = Vertx.vertx();
   return load(vertx).eventually(nothing -> vertx.close());
}
```

Advantage is that is introduces zero additional code & just a very little more test code.

Then in the `NeonBee` we just have to make sure, the temporary Vert.x instance is initialized after we set the logging stuff in the `create` method. I already spoke to @s4heid, is is anyways not nice that in `create()` we do static initialization here [2]. This results in that initialize multiple loggers when creating multiple NeonBee instances. Thus I would suggest, let's move the code of [2] to a "more general" place, keeping the new `create(Supplier<Future<Vertx>>, NeonBeeOptions, NeonBeeConfig)` signature of `create`. E.g.:

```java
static volatile Logger logger;

static void initialize(NeonBeeOptions options) {
    // old code ...

    if (logger === null) {
       synchronized(NeonBee.class) {
         if (logger === null) {
           // old logging initialization here
        }
      }
   }
}

// ...

static Future<NeonBee> create(Supplier<Future<Vertx>> vertxFutureSupplier, NeonBeeOptions options) {
   // static initializer, e.g. for logging
   initialize(options);

   return NeonBeeConfig.load().compose(config -> create(vertxFutureSupplier, options, config));
}

static Future<NeonBee> create(Supplier<Future<Vertx>> vertxFutureSupplier, NeonBeeOptions options,
            NeonBeeConfig config) {
   // static initializer, e.g. for logging
   initialize(options);
}
```

I am a little unsure about the `initialize` method, or if there is a nicer way to do this? However all in all, I think doing it with a own temporary Vert.x instance could spare us a lot of otherwise "dead" code. As we would have to clearly mark the new synchronous readYAML, readJSON, etc. methods as to "NEVER USE" anyways, because the asynchronous methods must be favored at all time anyways. What do you think?

If you say we keep the current synchronous way, I will finish my review of the existing code, if we say we switch to the temporary Vert.x instance instead, I would like to wait for @halber to do the change first.

Sorry for opening another round of discussion on this, even after @halber did invest a lot of work already, but sometimes you only start to see things clearly, if you see what impact is had on the code.

[1] https://vertx.io/docs/vertx-config/java/#_configuring_vert_x_itself
[2] https://github.com/SAP/neonbee/blob/main/src/main/java/io/neonbee/NeonBee.java#L218

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
halber: I have fixed this issue.

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
halber: I added the NeonBeePrometheusMeterRegistry class to filter for this registry. I also moved the default initialization of the PrometheusMeterRegistry from the NeonBeeConfig to the MetricsEndpoint. Only when the MetricsEndpoint is loaded, the default NeonBeePrometheusMeterRegistry is added to the CompositeMeterRegistry.

pk-work: Cool! I'm happy that we had a talk.

s4heid:
Hi @kristian,

thanks for your detailed feedback. There's a few things which came to my mind when reading your comment.

1. **blocking load methods**: In my opinion, loading the NeonBeeConfig upfront (blocking) makes sense in general as it is required before startup and we have to wait anyways. However, I agree with you that having the large amount of newly introduced blocking methods is not nice. Nobody else should use them, as they are only required for loading the config.

2. **initializing a temporary vertx**: I would like to avoid this for multiple reasons, if possible. In one of the experiments of @halber it turned out that it takes a while until the vertx is closed again. So either we have two vertx instances running at the same time (which throws a warning from vertx), or we have to wait until the this vertx instance is closed again (I'm not aware of the actual duration, though -> needs to be checked), which might impact performance. In the first case, I'd be concerned that we run into some side-effects of having multiple vertx instances at the same time.

3. **static initialization**: I think it is a good idea to do this. In order keep the change small we might do this in a separate change or refactor commit. If we decide to do it separately, I will create a follow-up issue.

---

After a discussion with @halber this morning, it seems that adding registries to the `CompositeMeterRegistry` is also possible during runtime. With this finding, we 
- don't have to introduce any blocking methods (1.) -> the changes for loading the `NeonBeeConfig` blocking can be reverted and we simply add the registries at a later point in time, 
- don't need the temporary vertx (2.), and 
- the `create()` method can remain as it is more or less.

@kristian If you don't mind, please wait with your review until the change has been adapted. I guess it will be quite different thereafter.

Best Regards,
Sebastian

kristian:
I agree to all your points @s4heid  and @halber.

In regards to 2., I get your point here. What Vert.x does is to close it's underlying Netty thread pools, which is why it takes some time to complete. As starting multiple Vert.x's side-by-side is actually not a problem at all and (if?) only closing Vert.x introduces the slowdown, we could also close the initial Vert.x instance "fire and forget" (most cleanly, we could just resolve the promise in our shutdown hook).

Generally I really do like to have loaded the config upfront, we will run into situations again, where we will require to have it upfront... so scrapping all of @halber 's work is maybe not the best idea (yet?). I really do like the idea, as well as the simplification in the method signatures introduced.

I just see the disadvantage of 1.: introducing one-purpose blocking methods, which should / must never to be used again. How to you like the fire & forget Vert.x close suggestion?

s4heid:
@kristian I like your suggestion and I think that your approach could be useful for other things as well. However, I am afraid  that it will blow up this change if we do not separate it into a different change / discussion. Therefore, I created #95. Feel encouraged to leave your comments there.

docs/metrics.md
```diff
@@ -0,0 +1,68 @@
...
+the `VertxOptions`. Additional `MeterRegistry` can be added to this `CompositeMeterRegistry` at runtime.
+
+The `MetricsEndpoint`, which by default provides the Prometheus metrics under the /metrics path, registers the
+`PrometheusMeterRegistry` when the `MetricsEndpoint` router is created.
+
+## configure additional metrics registries
+### when launching NeonBee
+To register own Micrometer MeterRegistry interface `MicrometerRegistryLoader` must be implemented
+and the implementing class must be specified in the configuration `io.neonbee.NeonBee.yaml` in the array `micrometerRegistries`
+array.
```
pk-work, line 25: 2 times array?

src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -75,6 +83,21 @@ public NeonBeeConfig(JsonObject json) {
         NeonBeeConfigConverter.fromJson(json, this);
     }
 
+    /**
+     * Load the default {@link PrometheusMeterRegistry} and try to load all {@link MicrometerRegistryLoader}s which are
```
pk-work, line 87: outdated comment?

src/main/java/io/neonbee/config/metrics/MicrometerRegistryLoader.java
```diff
@@ -0,0 +1,20 @@
...
+import io.micrometer.core.instrument.MeterRegistry;
+import io.vertx.micrometer.MicrometerMetricsOptions;
+
+/**
+ * Interface to implement if you want to add a {@link MeterRegistry} to the {@link MicrometerMetricsOptions}.
+ * <p>
+ * The implementing class MUST have a constructor without any arguments.
+ */
+@FunctionalInterface
+public interface MicrometerRegistryLoader {
```
pk-work, line 12: Why is this class required?

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    private static final LoggingFacade LOGGER = LoggingFacade.create();
+
+    private final String registryName;
+
+    /**
+     * Constructs a new instance of NeonBeePrometheusHandler.
+     *
+     * @param registryName The name of the micrometer registry
+     */
+    public NeonBeePrometheusHandler(String registryName) {
```
pk-work, line 29: Where I can configure the registryName? Haven't found something in server config.

src/test/java/io/neonbee/endpoint/metrics/NeonBeeMetricsIt.java
```diff
@@ -0,0 +1,117 @@
...
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.junit5.Checkpoint;
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+import io.vertx.micrometer.backends.BackendRegistries;
+
+@ExtendWith(VertxExtension.class)
+@SuppressWarnings("PMD.AvoidUnnecessaryTestClassesModifier")
+public class NeonBeeMetricsIt {
```
pk-work, line 37: What does "It" mean? IntegrationTest? We don't have used this pattern so far. When we want to introduce it, we should discuss it on Committer Meeting.

pk-work (changes requested):

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
halber: This can be configured like before throw the `MetricsEndpoint` configuration.

src/test/java/io/neonbee/endpoint/metrics/NeonBeeMetricsIt.java
halber: Yes it stands for IntegrationTest.

src/main/java/io/neonbee/config/metrics/MicrometerRegistryLoader.java
halber: We cannot load the MeterRegistry directly because sometimes it needs to be configured. For example, the DynatraceMeterRegistry requires some configuration. Of course, we could provide the Micrometer configuration map, but then we have to deal with the next problem that the MeterRegistry implementations do not all have the same constructor signature.

src/main/java/io/neonbee/endpoint/metrics/MetricsEndpoint.java
```diff
@@ -2,19 +2,39 @@
...
+    @SuppressWarnings("PMD.AvoidSynchronizedAtMethodLevel")
+    private static synchronized void addRegistry(Vertx vertx) {
+        CompositeMeterRegistry compositeMeterRegistry = NeonBee.get(vertx).getConfig().getCompositeMeterRegistry();
+        boolean isNotRegisterd = compositeMeterRegistry.getRegistries().stream()
+                .noneMatch(NeonBeePrometheusMeterRegistry.class::isInstance);
+        if (isNotRegisterd) {
+            compositeMeterRegistry.add(new NeonBeePrometheusMeterRegistry(PrometheusConfig.DEFAULT));
+        }
+    }
+
```
kristian, line 37: remove, the `NeonBeePrometheusMeterRegistry` should be set as default in the `NeonBeeConfig` (instead of initializing the `microMeterRegistries` with a `private List<String> registries = List.of();` initialize it with a `List.of("io.neonbee.endpoint.metric.NeonBeePrometheusMeterRegistry")`. This way it is clear (from a config perspective) which registies get added, it removes the requirement to (with some dubious instance variable) pass the composite metrics registry around via the `getConfig()` method and allows to not add the `NeonBeePrometheusMeterRegistry` if it is unwanted by the user. If there are no registries set in the config, the metrics endpoint should not get deployed in the `NeonBee` class (this way it can be disabled).

The downside of this is: How do you pass configuration from the config into the registry? I think later we should not only have `microMeterRegistries` be a `List<String>`, but add a new `MicroMeterRegistryConfig` object and make it a `List<MicroMeterRegistryConfig>`, that also allows to pass a JSON of options towards the registry implementation.

src/main/java/io/neonbee/endpoint/metrics/NeonBeePrometheusHandler.java
```diff
@@ -0,0 +1,63 @@
...
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.micrometer.backends.BackendRegistries;
+import io.vertx.micrometer.impl.PrometheusScrapingHandlerImpl;
+
+/**
+ * A Vert.x Web {@link io.vertx.ext.web.Route} handler for Prometheus metrics scraping.
+ * <p>
+ * The original Implementation doesn't work with {@link CompositeMeterRegistry}. This implementation fixes this.
+ */
+public class NeonBeePrometheusHandler extends PrometheusScrapingHandlerImpl {
```
kristian, line 19: I think we shall keep the original name. Call it:

PrometheusScrapingHandler (as it is in our NeonBee package, it is clear that it is ours).

Alternatively (please maybe quickly discuss / share with Pascal / Sebastian) call it:

NeonBeePrometheusScrapingHandler

src/main/java/io/neonbee/internal/helper/FileSystemHelper.java
```diff
@@ -104,7 +104,7 @@ private FileSystemHelper() {}
      *
      * @param vertx The related Vert.x instance
      * @param path  The path of the file
-     * @return a future to the converted {@link JsonObject}
+     * @return the {@link JsonObject}
```
kristian, line 107: keep the original comment, a {@link JsonObject} is no more correct

kristian (changes requested):

src/main/java/io/neonbee/NeonBee.java
```diff
@@ -241,7 +249,9 @@ public static NeonBee get(Vertx vertx) {
                 NeonBee neonBee = new NeonBee(vertx, options);
 
                 // load the configuration and boot it up, on failure close Vert.x
-                return neonBee.loadConfig().compose(config -> neonBee.boot()).recover(closeVertx).map(neonBee);
+                return neonBee.loadConfig().map(config -> config.setCompositeMeterRegistry(compositeMeterRegistry))
```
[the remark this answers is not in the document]
halber: Regarding point 1:
Unfortunately, we can't get rid of the setCompositeMeterRegistry method because we need the CompositeMeterRegistry in the VertxOptions and the NeonBeeConfig and it must be the same object.

src/main/java/io/neonbee/endpoint/metrics/MetricsEndpoint.java
halber: This is kind of the exact opposite of what I had discussed with @pk-work directly. The underlying discussion was this one here https://github.com/SAP/neonbee/pull/94#discussion_r778621669.

src/main/java/io/neonbee/NeonBee.java
```diff
@@ -217,12 +218,19 @@ public static NeonBee get(Vertx vertx) {
...
+        VertxOptions vertxOptions = new VertxOptions().setEventLoopPoolSize(options.getEventLoopPoolSize())
+                .setWorkerPoolSize(options.getWorkerPoolSize());
+
+        CompositeMeterRegistry compositeMeterRegistry = new CompositeMeterRegistry();
+        vertxOptions.setMetricsOptions(
+                new MicrometerMetricsOptions().setMicrometerRegistry(compositeMeterRegistry).setEnabled(true));
+
         // create a Vert.x instance (clustered or unclustered)
-        return vertxFutureSupplier.get().compose(vertx -> {
+        return vertxFutureSupplier.apply(vertxOptions).compose(vertx -> {
```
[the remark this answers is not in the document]
halber: Maybe we can change this if we [Load NeonBee options upfront](https://github.com/SAP/neonbee/issues/95). Currently I have to set the same CompositeMeterRegistry in the VertxOptions and in the NeonBeeConfig. If we had the NeonBeeConfig before we create the Vertx object, then I could easily change this.

src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -49,8 +52,12 @@
 
     private String timeZone = DEFAULT_TIME_ZONE;
 
+    private List<String> registries = List.of();
+
+    private CompositeMeterRegistry compositeMeterRegistry;
```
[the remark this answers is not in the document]
halber: Maybe I missed something, but I don't see how I can delete this.

kristian:
@halber I will now look into your comments. sorry for the delay. let us close this one soon :)

kristian (approved):

kristian:
lgtm! 👍

pk-work (approved):

PR #97, opened 2022-02-01
fix: null values queried with contains

halber:
This fixes the problem that if an entity attribute can contain null values, it was not possible to query this attribute with a query like contains(x,'y') or contains(x,'Y').


kristian (approved):
lgtm! thanks @halber, feel free to merge

PR #100, opened 2022-02-17
fix: NPE ImmutableJsonObject/Array for null values

kristian:


src/main/java/io/neonbee/internal/json/ImmutableJsonArray.java
```diff
@@ -66,12 +66,14 @@ public ImmutableJsonArray(JsonArray arr) {
...
-        return new ImmutableJsonObject(super.getJsonObject(pos));
+        JsonObject object = super.getJsonObject(pos);
+        return object != null ? new ImmutableJsonObject(object) : null;
     }
 
     @Override
     public ImmutableJsonArray getJsonArray(int pos) {
-        return new ImmutableJsonArray(super.getJsonArray(pos));
+        JsonArray array = super.getJsonArray(pos);
+        return array != null ? new ImmutableJsonArray() : null;
```
halber, line 76:
```suggestion
        return array != null ? new ImmutableJsonArray(array) : null;
```

halber (changes requested):
please fix

PR #104, opened 2022-02-23
fix: set ALL as default for active profiles

pk-work:


pk-work:
This could also be solved in Launcher line 49 by simply adding

```java
CLIConfigurator.inject(commandLine, options);
if (options.getActiveProfiles()) {
options.setActiveProfiles(Set.of(All);
```
But then NeonBeeOptions would only defaults to ALL when it is "generated" by Launcher.

kristian:
There is no debate that imho that `ALL` should be the default, so line:

```java
private Set<NeonBeeProfile> activeProfiles = Set.of(NeonBeeProfile.ALL);
```

Is absolutely necessary. Question for me is: Is setting _no_ active profile (so purposefully calling `setActiveProfile()` a use case, or not? Does it make sense to set an empty profile, e.g. to only deploy system verticles? This way all other lines than the line above, would not require to be changed.

Also if they be changed, I don't think a change to the setter is necessary, because you do already check it it the getter, so I would suggest (if we decide to keep the logic) to change the getter to:

```java
return activeProfiles == null || activeProfiles.isEmpty() ? Set.of(NeonBeeProfile.ALL) : activeProfiles;
```

What do you think?

pk-work:
Would be fine for me.

kristian:
Ok, sounds good. Thanks @pk-work

pk-work:
Just to double check:
 - We say the use case of an empty list of profiles is not required.
 - We say passing an empty list to setProfiles should have the same behavior than  passing `ALL`.

If this is fine we only need to change the line the getter to:
`return activeProfiles.isEmpty() ? Set.of(NeonBeeProfile.ALL) : activeProfiles;`

Because activeProfiles can't be null, if it is initialized with a Set.

kristian:
No, I would say we should decide for either / or:

- Default should be ALL
- Can be overridden with an empty list, meaning no profiles

So I would only suggest to set the ALL profile as default

pk-work:
This will not work, because the CLI in Launcher (when no profiles passed), will always call setActiveProfiles with an empty array which results in an empty List. So it will override the default value from the initialization.

Then we would need to change the Launcher to add the default value again, if no profiles argument is passed.

kristian:
I did enable the `EnvironmentAwareCommandLine` to respect default options also for multi-valued options, see PR #106, this will also allow to set an empty list of active profiles. What do you think?

kristian:
Fixed with PR#106

PR #105, opened 2022-02-24
ci: automate dependency upgrade (of vertx)

s4heid:
With the dependency upgrade workflow, bumping the version of vertx (and
in the future also other dependencies) will be automated.

The workflow checks every day at midnight if a new release version of
vertx is available. If it finds a new version, a pull request will be
created which updates the version in the `build.gradle`.

### DIFF
--- a/.github/workflows/.auto-upgrade.json
+++ b/.github/workflows/.auto-upgrade.json
@@ -1,0 +1,7 @@
+[
+    {
+        "name": "vertx",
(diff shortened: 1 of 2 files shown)


s4heid:
For reference, here is an example PR which was generated in my personal fork: https://github.com/s4heid/neonbee/pull/1

halber (approved):

PR #106, opened 2022-03-01
fix: allow no active profile, make ALL the default

kristian:
Allow to set null / an empty list as active profiles, which means that no profiles are to be deployed and NeonBee will start with the system verticles only.


kristian:
The CLA assistant is having some kind of hick-up again, the change is green though!

kristian:
Maybe @pk-work you need to also (again) sign the CLA, because it is a co-signed commit?!

pk-work (approved):

PR #108, opened 2022-03-07
refactor: improve wildcard handling in SelfFirstClassLoader

kristian:
So far the SelfFirstClassLoader only allowed to specify parentPreferred className if the wildcard was at the end of the string like:

io.neonbee.*

If the wildcard was in any other position, the check failed. In additional the general implementation of the check was implemented inefficiently, because of a list traversal, on every check, which is, for modules, every class that is to be loaded.

This change improves the wildcard logic, by compiling a pattern on initialization of the SelfFirstClassLoader, also allowing more complex patterns like:

io.neonbee.*.TestClass

### DIFF
--- a/src/main/java/io/neonbee/config/NeonBeeConfig.java
+++ b/src/main/java/io/neonbee/config/NeonBeeConfig.java
@@ -51,7 +51,7 @@ public class NeonBeeConfig {
 
     private String trackingDataHandlingStrategy = DEFAULT_TRACKING_DATA_HANDLING_STRATEGY;
 
-    private List<String> platformClasses = List.of();
+    private List<String> platformClasses = List.of("io.vertx.*", "io.neonbee.*", "org.slf4j.*", "org.apache.olingo.*");
 
     private String timeZone = DEFAULT_TIME_ZONE;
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/neonbee/config/NeonBeeConfig.java
```diff
@@ -51,7 +51,7 @@
 
     private String trackingDataHandlingStrategy = DEFAULT_TRACKING_DATA_HANDLING_STRATEGY;
 
-    private List<String> platformClasses = List.of();
+    private List<String> platformClasses = List.of("io.vertx.*", "io.neonbee.*", "org.slf4j.*", "org.apache.olingo.*");
```
pk-work, line 54: What exactly is the idea of the prefilled platform classes? Because this will not match ALL platform classes .. Just a best guess? Then I would leave it empty and add these values in the example config, which would be used by the most interested people by default.

kristian: This is why the default makes more sense than just providing it as an example... If somebody wants to do it differently, it has to be a pretty specific use case.

pk-work: lgtm

PR #113, opened 2022-03-11
feat: add metrics to DataVerticle

halber:
This feature adds metrics for the DataVertical methods requestData and retrieveData. Four metrics are implemented.

| metric | discription |
|---|---|
| Time metric | How long it takes to retrieve the data. |
| Status counter | How many requests completed successfully, how many requests failed |
| Number of requests | Total of all requests |
| Active requests | How many requests are currently open |

The metrics are disabled by default. You can enable the metrics for all DataVerticle by setting the NeonBeeConfig metrics.enabled value to true.
io.neonbee.NeonBee.yaml:
```yaml
# Other settings omitted.
metrics:
    enabled: true
```

The metrics can be also configured using the verticle configuration. To enable the metrics for a individual DataVerticle, it is necessary to add the config.metrics.enabled key with the value true. In addition, you can enable only specific metrics by specifying the metrics configuration name and a true value in the verticle configuration. If you have added a metrics configuration key, only the values with the specified true value are enabled. All others are disabled.

Complete yaml example configuration to enable the metrics:
```yaml
config:
  metrics:
    enabled: true
    meterRegistryName: SomeRegistryName
    reportNumberOfRequests: true
    reportActiveRequests: true
    reportStatusCounter: true
    reportTiming: true
```

### DIFF
--- a/src/main/java/io/neonbee/data/DataVerticle.java
+++ b/src/main/java/io/neonbee/data/DataVerticle.java
@@ -15,8 +15,10 @@ import static io.vertx.core.Future.succeededFuture;
 import static java.util.Collections.emptyList;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
@@ -29,10 +31,15 @@ import org.apache.olingo.commons.api.edm.FullQualifiedName;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 
+import io.micrometer.core.instrument.ImmutableTag;
+import io.micrometer.core.instrument.Tag;
 import io.neonbee.NeonBee;
 import io.neonbee.NeonBeeDeployable;
+import io.neonbee.config.MetricsConfig;
 import io.neonbee.data.DataRequest.ResolutionStrategy;
 import io.neonbee.data.internal.DataContextImpl;
+import io.neonbee.data.internal.metrics.ConfiguredDataVerticleMetrics;
+import io.neonbee.data.internal.metrics.DataVerticleMetrics;
 import io.neonbee.internal.helper.FunctionalHelper;
 import io.neonbee.logging.LoggingFacade;
 import io.vertx.core.AbstractVerticle;
@@ -46,21 +53,32 @@ import io.vertx.core.Vertx;
 import io.vertx.core.eventbus.DeliveryOptions;
 import io.vertx.core.eventbus.MessageCodec;
 import io.vertx.core.eventbus.ReplyException;
+import io.vertx.core.json.JsonObject;
 
+@SuppressWarnings("PMD.GodClass")
 public abstract class DataVerticle<T> extends AbstractVerticle implements DataAdapter<T> {
     /**
      * The name of the context header of an event bus message.
      */
     public static final String CONTEXT_HEADER = "context";
 
+    /**
+     * Metrics configuration name.
+     */
+    public static final String CONFIG_METRICS_KEY = "metrics";
+
     static final String RESOLUTION_STRATEGY_HEADER = "resolutionStrategy";
 
     private static final LoggingFacade LOGGER = LoggingFacade.create();
+
+    private static final String SUCCEEDED_RESPONSE_COUNT = "succeeded response count";
 
     @SuppressWarnings("UnnecessaryLambda") // overridden in DummyVerticleHelper, as getNamespace is final
     private final Supplier<String> namespaceSupplier =
             () -> Optional.ofNullable(this.getClass().getAnnotation(NeonBeeDeployable.class))
                     .map(NeonBeeDeployable::namespace).map(Strings::emptyToNull).map(String::toLowerCase).orElse(null);
+
+    private DataVerticleMetrics dataVerticleMetrics;
 
     /**
      * The name of this data verticle (must be unique in one cluster)
@@ -112,6 +130,8 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
     @Override
     public void init(Vertx vertx, Context context) {
         super.init(vertx, context);
+        JsonObject metrics = getMetricsConfig(NeonBee.get().getConfig().getMetricsConfig());
+        this.dataVerticleMetrics = ConfiguredDataVerticleMetrics.configureMetricsReporting(metrics);
 
         // if present, register the custom codec. IMPORTANT: do NOT register the codec in the start method, as the
         // codec will need to be available on all instances, even if no instance of the verticle is started later on
@@ -125,6 +145,27 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
                 }
             }
         }
+    }
+
+    /**
+     * Get the metric configuration.
+     *
+     * @param globalMetricsConfig the global {@link MetricsConfig} object.
+     * @return the metric configuration as JSON object.
+     */
+    @VisibleForTesting
+    JsonObject getMetricsConfig(MetricsConfig globalMetricsConfig) {
+        MetricsConfig metricsConfig =
+                globalMetricsConfig == null ? new MetricsConfig().setEnabled(false) : globalMetricsConfig;
+        JsonObject verticleMetricsConfig =
+                config() == null ? new JsonObject() : config().getJsonObject(CONFIG_METRICS_KEY, new JsonObject());
+
+        Boolean enabled = verticleMetricsConfig.getBoolean(ConfiguredDataVerticleMetrics.ENABLED);
+        // if there is no setting for the metric enabled in the DataVerticle configuration use the global setting.
+        if (enabled == null || Boolean.TRUE.equals(enabled)) {
+            verticleMetricsConfig.put(ConfiguredDataVerticleMetrics.ENABLED, metricsConfig.isEnabled());
+        }
+        return verticleMetricsConfig;
     }
 
     /**
@@ -262,7 +303,9 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
     public <U> Future<U> requestData(DataRequest request, DataContext context) {
         LOGGER.correlateWith(context).debug("Data verticle {} requesting data from {}", getQualifiedName(), request);
 
-        return requestData(vertx, request, context);
+        Future<U> future = requestData(vertx, request, context);
+        reportRequestDataMetrics(request, future);
+        return future;
     }
 
     /**
@@ -277,6 +320,7 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
      */
     public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataContext context) {
         DataSource<?> dataSource = request.getDataSource();
+
         if (dataSource != null) {
             return dataSource.retrieveData(request.getQuery(), context).map(FunctionalHelper::uncheckedMapper);
         }
@@ -321,6 +365,24 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
         }
 
         return failedFuture(new IllegalArgumentException("Data request did not specify what data to request"));
+    }
+
+    private <U> void reportRequestDataMetrics(DataRequest request, Future<U> future) {
+        List<Tag> tags;
+        if (request.getQuery() == null || request.getQuery().getQuery() == null) {
+            tags = List.of();
+        } else {
+            tags = List.of(new ImmutableTag("query", request.getQuery().getQuery()));
+        }
+        String qualifiedName = request.getQualifiedName();
+
+        dataVerticleMetrics.reportTimingMetric("request.data.timer." + qualifiedName, "time to retrieve the data", tags,
+                future);
+        dataVerticleMetrics.reportStatusCounter("request.data.counter." + qualifiedName, SUCCEEDED_RESPONSE_COUNT, tags,
+                future);
+        dataVerticleMetrics.reportActiveRequestsGauge("request.data.active.requests." + qualifiedName,
+                "Number of requests waiting for a response", List.of(), future);
+        dataVerticleMetrics.reportNumberOfRequests("request.counter." + qualifiedName, "Number of requests sent", tags);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


src/test/java/io/neonbee/data/internal/metrics/DataVerticleMetricsImplTest.java
```diff
@@ -0,0 +1,110 @@
...
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+
+@ExtendWith(VertxExtension.class)
+class DataVerticleMetricsImplTest {
+
+    @Test
+    @Timeout(value = 1, timeUnit = TimeUnit.MINUTES)
+    void testMetircsOnPrometheusEndpoint(Vertx vertx, VertxTestContext context) throws Exception {
```
s4heid, line 32:
```suggestion
    void testMetricsOnPrometheusEndpoint(Vertx vertx, VertxTestContext context) throws Exception {
```

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -490,20 +527,53 @@ private ResolutionRoutine resolutionRoutineForStrategy(ResolutionStrategy strate
...
+            tags.add(new ImmutableTag("name", name));
+        }
+        String namespace = getNamespace();
+        if (namespace != null) {
+            tags.add(new ImmutableTag("namespace", namespace));
+        }
+        return tags;
+    }
+
+    private void reportretRieveDataMetrics(List<Tag> tags, Future<T> future) {
```
s4heid, line 567:
```suggestion
    private void reportRetrieveDataMetrics(List<Tag> tags, Future<T> future) {
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,156 @@
...
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.neonbee.logging.LoggingFacade;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+import io.vertx.micrometer.MicrometerMetricsOptions;
+import io.vertx.micrometer.backends.BackendRegistries;
+
+/**
+ * This class is for configure the metrics to report.
```
s4heid, line 16:
```suggestion
 * This class configures the metrics to report for a {@link DataVerticle}.
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public static final String ENABLED = "enabled";
+
+    /**
+     * Key for the name of the registry to use.
+     */
+    public static final String METER_REGISTRY_NAME = "meterRegistryName";
+
+    /**
+     * Key for reporting request count.
```
s4heid, line 31:
```suggestion
     * Key for reporting the request count.
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,156 @@
...
+     */
+    public static final String ACTIVE_REQUESTS = "reportActiveRequests";
+
+    /**
+     * Key for reporting the status counter.
+     */
+    public static final String STATUS_COUNTER = "reportStatusCounter";
+
+    /**
+     * Key for reporting timming values.
```
s4heid, line 46:
```suggestion
     * Key for reporting timing values.
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,156 @@
...
+                            Boolean.TRUE.equals(metricsConfig.getBoolean(NUMBER_OF_REQUESTS)) ? metricsImpl
+                                    : DUMMY_IMPL,
+                            Boolean.TRUE.equals(metricsConfig.getBoolean(ACTIVE_REQUESTS)) ? metricsImpl : DUMMY_IMPL,
+                            Boolean.TRUE.equals(metricsConfig.getBoolean(STATUS_COUNTER)) ? metricsImpl : DUMMY_IMPL,
+                            Boolean.TRUE.equals(metricsConfig.getBoolean(TIMING)) ? metricsImpl : DUMMY_IMPL);
+                }
+            }
+        } else {
+            return DUMMY_IMPL;
+        }
```
s4heid, line 134: nit; I think it would be nicer to read if we would do an early return here, and get rid of one level of nested `if`'s.

```suggestion
        if (metricsConfig == null && Boolean.TRUE.equals(metricsConfig.getBoolean(ENABLED))) {
            return DUMMY_IMPL;
        }

            String meterRegistryName =
                    metricsConfig.getString(METER_REGISTRY_NAME, MicrometerMetricsOptions.DEFAULT_REGISTRY_NAME);
            MeterRegistry registry = BackendRegistries.getNow(meterRegistryName);

            if (registry == null) {
                LOGGER.error(
                        "Micrometer registry hasn't been registered yet or it has been stopped. Metrics will not be sent.");
                return DUMMY_IMPL;
            } else {
                DataVerticleMetrics metricsImpl = new DataVerticleMetricsImpl(registry);

                int fieldNameSize = metricsConfig.containsKey(METER_REGISTRY_NAME) ? 2 : 1;
                boolean activateAllMetrics = metricsConfig.size() == fieldNameSize;
                if (activateAllMetrics) {
                    return metricsImpl;
                }
                return new ConfiguredDataVerticleMetrics(
                        Boolean.TRUE.equals(metricsConfig.getBoolean(NUMBER_OF_REQUESTS)) ? metricsImpl : DUMMY_IMPL,
                        Boolean.TRUE.equals(metricsConfig.getBoolean(ACTIVE_REQUESTS)) ? metricsImpl : DUMMY_IMPL,
                        Boolean.TRUE.equals(metricsConfig.getBoolean(STATUS_COUNTER)) ? metricsImpl : DUMMY_IMPL,
                        Boolean.TRUE.equals(metricsConfig.getBoolean(TIMING)) ? metricsImpl : DUMMY_IMPL);
            }
        }
```

src/main/java/io/neonbee/data/internal/metrics/DataVerticleMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param name        the name of the metric
+     * @param description description of the metric
+     * @param tags        dimensions of a meter used to classify the metric
+     * @param future      the future to measure
+     */
+    void reportActiveRequestsGauge(String name, String description, List<Tag> tags, Future<?> future);
+
+    /**
+     * Report status metrics.
```
s4heid, line 30:
```suggestion
     * Reports status metrics.
```

s4heid (changes requested):

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
halber: I agree, it is nicer to read, but I disagree with your suggested code. You forgot to invert the Boolean.TRUE.equals(metricsConfig.getBoolean(ENABLED)) part.


```
if (metricsConfig == null || Boolean.FALSE.equals(metricsConfig.getBoolean(ENABLED, Boolean.FALSE))) {
    return DUMMY_IMPL;
}
```

halber: My suggestion is also wrong. It has to be like this:

```
if (metricsConfig == null || !Boolean.TRUE.equals(metricsConfig.getBoolean(ENABLED))) {
    return DUMMY_IMPL;
}
```

s4heid:
@halber it would be nice if we could extend the [docs](https://github.com/SAP/neonbee/blob/main/docs/metrics.md) with these additional configuration parameters which you already summarized in the description of this PR. 

Just realized that the configuration options you provide (e.g. enabling/disabling metrics, enabling/disabling reporting of certain checks) should be done similarly in #118 for health checks, such that the NeonBee user has a better user experience. In order to have similar configuration options for health and metrics, maybe we can discuss the "format" in the upcoming NeonBee committer meeting.

Apart from that, I am fine with the change.

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -127,6 +146,23 @@ public void init(Vertx vertx, Context context) {
...
+     * @return the metric configuration as JSON object.
+     */
+    @VisibleForTesting
+    JsonObject getMetricsConfig(MetricsConfig metricsConfig) {
+        JsonObject metrics = config().getJsonObject(CONFIG_METRICS_KEY, new JsonObject());
+        Boolean enabled = metrics.getBoolean(ConfiguredDataVerticleMetrics.ENABLED);
+        // if there is no setting for the metric enabled in the DataVerticle configuration use the global setting.
+        if (enabled == null || Boolean.TRUE.equals(enabled)) {
+            metrics.put(ConfiguredDataVerticleMetrics.ENABLED, metricsConfig.isEnabled());
+        }
```
s4heid, line 162:
```suggestion
        JsonObject metrics = config().getJsonObject(CONFIG_METRICS_KEY, new JsonObject());
        boolean enabled = metrics.getBoolean(ConfiguredDataVerticleMetrics.ENABLED, true);
        // if there is no setting for the metric enabled in the DataVerticle configuration use the global setting.
        if (enabled) {
            metrics.put(ConfiguredDataVerticleMetrics.ENABLED, metricsConfig.isEnabled());
        }
```

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -490,20 +547,53 @@ public Future<T> execute(DataQuery query, DataContext context) {
...
+                        })).map(Future.class::cast).collect(Collectors.toList())).otherwiseEmpty();
             }).compose(requiredCompositeOrNothing -> {
+                List<Tag> tags = retrieveDataTags();
                 try {
-                    return retrieveData(query, new DataMap(requestResults), context);
+                    Future<T> future = retrieveData(query, new DataMap(requestResults), context);
+                    reportRetrieveDataMetrics(tags, future);
+                    return future;
                 } catch (Exception e) {
+                    dataVerticleMetrics.reportStatusCounter("retrieve_data_counter_" + getAddress(),
```
s4heid, line 562: is this actually the right format? shouldn't it be:

```suggestion
                    dataVerticleMetrics.reportStatusCounter("retrieve.data.counter." + getAddress(),
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,160 @@
...
+        this.reportNumberOfRequests = reportNumberOfRequests;
+        this.reportActiveRequestsGauge = reportActiveRequestsGauge;
+        this.reportStatusCounter = reportStatusCounter;
+        this.reportTimingMetric = reportTimingMetric;
+    }
+
+    /**
+     * Configure the {@link DataVerticleMetrics} to report.
+     *
+     * To enable all metrics, you must specify the enabled key with the value ture. Otherwise, no metrics will be
```
s4heid, line 80:
```suggestion
     * To enable all metrics, you must specify the "enabled" key with the value true. Otherwise, no metrics will be
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,160 @@
...
+     *     "reportTiming" : true
+     * }
+     * }
+     * </pre>
+     *
+     * @param metricsConfig {@link JsonObject} containing the metrics configuration.
+     * @return configured {@link DataVerticleMetrics}
+     */
+    public static DataVerticleMetrics configureMetricsReporting(JsonObject metricsConfig) {
+        if (metricsConfig == null || !Boolean.TRUE.equals(metricsConfig.getBoolean(ENABLED))) {
```
s4heid, line 107:
```suggestion
        if (metricsConfig == null || !metricsConfig.getBoolean(ENABLED, false)) {
```

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
```diff
@@ -0,0 +1,160 @@
...
+    private static DataVerticleMetrics configureDataVericleMetrics(JsonObject metricsConfig,
+            DataVerticleMetrics metricsImpl) {
+
+        int fieldNameSize = metricsConfig.containsKey(METER_REGISTRY_NAME) ? 2 : 1;
+        boolean activateAllMetrics = metricsConfig.size() == fieldNameSize;
+        if (activateAllMetrics) {
+            return metricsImpl;
+        } else {
+            return new ConfiguredDataVerticleMetrics(
+                    Boolean.TRUE.equals(metricsConfig.getBoolean(NUMBER_OF_REQUESTS)) ? metricsImpl : DUMMY_IMPL,
```
s4heid, line 134:
```suggestion
                    metricsConfig.getBoolean(NUMBER_OF_REQUESTS, false) ? metricsImpl : DUMMY_IMPL,
```

src/main/java/io/neonbee/data/internal/metrics/DataVerticleMetrics.java
```diff
@@ -0,0 +1,50 @@
...
+
+import io.micrometer.core.instrument.Tag;
+import io.vertx.core.Future;
+
+public interface DataVerticleMetrics {
+
+    /**
+     * Reports the sum of all calls.
+     *
+     * @param name        the name of the metric
```
s4heid, line 13:
```suggestion
     * @param name        name of the metric
```

for all, to make param description consistent

src/main/java/io/neonbee/data/internal/metrics/DummyDataVerticleMetrics.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.List;
+
+import io.micrometer.core.instrument.Tag;
+import io.vertx.core.Future;
+
+/**
+ * This implementation of the {@link DataVerticleMetrics} interface is used when metrics are disabled.
+ */
+public class DummyDataVerticleMetrics implements DataVerticleMetrics {
```
s4heid, line 11: What do you think about renaming this to NoopDataVerticleMetrics as this class is actually doing null operations. Dummy sounds more like it's doing something "simple", which is not true.

```suggestion
public class NoopDataVerticleMetrics implements DataVerticleMetrics {
```

halber: renamed

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
halber: With your suggestion the case that the key is present but with a null value would cause an NPE.

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
s4heid: No, because you are already checking if `metricsConfig` is null. The latter part is not evaluated.

halber: No, if the JSON object metricsConfig would look like this:
`{
  "enabled": null
}`
then we get a NullPointerException with your suggested code change.

I even wrote a JuntiTest for the case: io.neonbee.data.internal.metrics.ConfiguredDataVerticleMetricsTest#disabledNull

src/main/java/io/neonbee/data/DataVerticle.java
halber: In your case I would get a NullPointerExcpetion if the ConfiguredDataVerticleMetrics.ENABLED key has the value null.

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -127,6 +146,23 @@ public void init(Vertx vertx, Context context) {
...
     }
 
+    /**
+     * Get the metric configuration.
+     *
+     * @param metricsConfig the global {@link MetricsConfig} object.
+     * @return the metric configuration as JSON object.
+     */
+    @VisibleForTesting
+    JsonObject getMetricsConfig(MetricsConfig metricsConfig) {
```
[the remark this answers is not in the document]
halber: I will change that.

src/main/java/io/neonbee/data/internal/metrics/ConfiguredDataVerticleMetrics.java
s4heid: ah, you are right :) consider my comment as invalid

s4heid (approved):

PR #116, opened 2022-04-09
feat(plugin): add initial gradle plugin

pk-work:


pk-work:
I will add a commit with a verticles plugin in the next days to this PR

kristian:
Thanks a lot @pk-work, I really do think this plugin will much simplify and enable usage of NeonBee modules thanks for all the work!

FYI: (no real review just had a quick look), imho this implementation of the plugin is still too big and I do see many flaws, that I think atm. prevent merging ...

First and foremost: I don't see why we should be making so many "assumptions" about Gradle projects... we have and should adhere to standards, because I do think, this plugin will actually break any existing Java project / structure. Here are some examples what I mean:

- The whole QualityPlugin: There are quality plugins out there and why should we (NeonBee) maintain our own "plugin-of-plugins" of quality plugins out there? We are not responsible for other's project's quality.
- The ApplicationExtension, defining dependencies through extensions is generally not very well regarded in Gradle... this is what Gradles dependency handler is for, not extensions
- Stuff that changes project configuration, like `project.archivesBaseName = config.getComponentName()`, a plugin should never do that... how would the user know that we are messing up his project? (Also what is a "component name"? I don't think we should introduce new terminology to Gradle.
- Defining own publications with our own naming scheme is not ok... We must not influence if anyone is having publications in any name... MavenPublication is just one type of many publication types Gradle supports, "maven" as the name of the publication is _our_ publication name, but the user may choose an own one
- Creating publications with "artifact source: new LazyPublishArtifact(project.tasks.named(ModulePlugin.MODELS_MODULE_JAR_TASK_NAME))" is not standard compliant by any means. Gradle works with components for publishing. Artifacts should never be added to publications manually any longer, it will mess up Gradles own gradle-metadata.xml handling (e.g. if you are using test fixtures)
- Smaller things like, using the file name to differentiate between the type of module `it.getName().endsWith('-module.jar') || it.getName().endsWith('-models.jar')` is _not_ a right use. Gradle supports "usage types" for that purpose. A file can be named whatever and a classifier is generally only to be used to differentiate between different execution environments. Thus, yes we can use the classifier for `-module` but definitely _not_ for `-models`. Models must _not_ be the same artifact to publish. One artifact has one pom.xml. Meaning if you would make the `-models` part of the same module identifier (in the maven sense), so say com.example:example-neonbee-module and in there are two artifacts with different classifiers, like: `a.jar` and `a-models.jar` you essentially say: regardless if you consume `a.jar` or `a-models.jar` they are basically "the same", but with a different packaging and both share for example the same dependency tree that is defined in pom.xml. Also comparing the name of the JAR file is never a safe indication to determine the classifier. Every segment in JARs is separated by -. You cannot tell if it is a classifier or some other part of the file name.
- Copying around JAR files (to put them into the /verticles directory for running) is unnecessary and error prone. After dependency resolution the file is somewhere on the disk already. You should use the -mjp option to just refer to the modules to get loaded by NeonBee without copying them around. What if the directory isn't clean, because the file was renamed? The directory is not cleared or anything... that is just one thing.

I don't want to be too hard, so please don't take my statements to hard... 👍 In my humble opinion we should really try to keep the plugin as minimal as possible:

- One task for building a model and adding it to a configuration (nothing more, no assumption, just "compileModels" and create a consumable (not resolvable) configuration, that you can (with Gradles dependency / configuration logic) add to where ever you need it
- One task for running NeonBee and defining modules as "inputs" (this is where you can _manually_ link in the configuration from the previous step, if you desire to, no more)
- One task (much similar to the ShadowPlugin) that just registers one new build task for building module... Adding it to a publication, naming it, defining properties to other JARs or whatever, is the users responsibility.

So in total, three "little-ish" tasks, that can all go into a "NeonBeePlugin". That minimalistic plugin concept, much more complies to Gradles principles, which is "making no assumptions". Again thanks a lot for all the efforts Pascal and I hope you don't loose motivaion on this one 👍 This will be great!

kristian (changes requested):
See comment in conversation. I don't think the way this plugin is structured should be the way we go with it.

kristian:
Some more general thoughts:

**"Module" Terminology:** now that we deviate into the Gradle space, we are slowly, but surely facing the problem that our term "module" for essentially a FatJar with some extra MANIFEST.MF attributes is a heavily overloaded terminology in our space right now:

- Java has introduced "module" loading concept with Jigsaw and JSR 376 which defines the Java Platform "Module" System. Javas Modules are something different than our modules.
- Gradle uses the term "module" all over it's dependency resolution mechanism to refer to "the thing behind a external dependency", so e.g. ExternalDependencies are identified by a "moduleIdentifier". So our module is then for Gradle a "module-module" essentially which is not nice...

Thus we have to be very clear in our communication in Gradle and always call it "nbmodule" or similar (see the naming convention guide).

To fuel the discussion I would even like to make the following assumption: We should (outside of NeonBee) never talk about "modules"... the output of our plugin is and should be a JAR... a JAR that has some special attributes so we refer to it as a NeonBee Module for that reason... however for the end-user it is a normal JAR, that he happens to be able to deploy "seamlessly" to a NeonBee server. I don't think we have to put in any "preventive measure" here! We are a "plain old JAR"... if the user wants to use that JAR somewhere else OK! If the user loads a "non-Module-JAR" in NeonBee, well simply nothing happens, so OK! The user will know that he, if he wants to create a distributable artifact for NeonBee will publish it, e.g. as "MyCoolSever-nb.jar" or something... Vert.x plugins can also only be used in Vert.x, do they define a own JAR format? Of cause they don't... does everbody who consumes Vert.x plugin know that they are intended to be used with Vert.x? Yes! Sure, because the developer who created it released it as "vertx-plugin-dofancysstuff"! So why should we bother how the stuff our plugin creates is used? So also here in my oppinion we should keep it as minimal as possible. Add the MANIFEST.MF entries, maybe a "nbmodule-version" identifier, maybe (!) create a semi-fat JAR, that's it. Rest is up to the user to... this already leads me to the second general topic I want to bring up:

**Most Lightweight Plugin (MLP):** After thinking about it even a little more, this whole plugin(s) raise my general concern: Why do we have so many plugins for such little tasks? Generally I first thought we need three, now I even narrowed it down to two tasks only! Hear me out, this is how the plugin should be connected:

We need one task that:

- Takes "models" and / or
- Other (also "nbmodule") JARs as an input (similar or equal to the ShadowJar task)
- It outputs a "nbmodule JAR" by consolidating the inputs, also other "nbmodules" that are passed into should get consolidated
- It should provide possibilities to configure what is put into the "FatJar" to work together with NeonBee's `SelfFirstClassLoader` concept. So what we *should* create (in comparison go the ShadowJarPlugin) is only a "semi-fat-jar", so no "uber-jar", but a "middle-jar" ;). So what I even would assume and make the case is that: What the user wants is actually to define a self-first class loader whitelist in Gradle which is then used "inverse" for compiling the module (so everthing that is on the list should be excluded of the "semi-fat-jar") and everything that is on the list should get loaded from the NB classloader when executing (becomes relevant later).

Now we can decide if there should be a task that compiles models, as in what should it compile it? Because the above task can already be used to compile it into a "nbmodule" JAR... Shouln't it be enough to compile a given model to an (otherwise empty) nbmodule JAR? So we have essentially three types of module JARs:

1. those with code / verticles
2. those with models only
3. a composition of any combination of the above

So the only second task that we need in my oppionion is a specialized `JavaExec` task (not an own application plugin, because there is an application plugin already! It's for Java applications... we are a Java application and I don't think we *should* be a particularly special one!). So this second task has only three inputs and no outputs"

- Again "models" (if we want to support supplying them into)
- nbmodule JARs that get loaded using -mjp and
- The before mentioned configuration that defines the `SelfFirstClassloader` whitelist

With that the user can do everything he wants to so: Compile models, compile nbmodules and run neonbee easier. That's it.

pk-work:
Hi @kristian,

thank you for your "Quick Look Review". I know you are busy at the moment so I really appreciate you taking the time.

I agree with many of the concerns you have raised, especially those related to poor Gradle coding. I already knew some of the problems when I typed the lines of code, but I couldn't figure out a better way to solve it. I would need help to fix it.

I'm also happy, that this PR led to the more general question about "Module" Terminology. But for now I'd like to concentrate on the Plugin.

The main goal of this plugin is to make NeonBee easier to consume.

- **Lower the barrier to entry:** For now there is no way to simply create a new Gradle Project, write a Verticle and start / deplo it with NeonBee. With the NeonBee Application Plugin you don't need to take care about anything, just write you Verticle and see the results. This is is the foundation for the "Getting Started" Guide I'd like to write in the Future.

Does an Application Plugin already exist? Of course, I'm using it, but the NeonBee Application Plugin is a preconfigured version of it, to lower the barrier to entry.

- **The Quality Plugin:** You should not use the NeonBee Quality Plugin if you only want to have some quality checks in your project. You should use it, when you want to have the same rules for quality as in NeonBee. I see a big potential here. When we start to develop NeonBee extensions (like vertx-web), we can easily apply the same rules there.

And maybe it is also a nice feature for everyone who don't want to get familiar with all of these plugins and how to configure them ...

- **The Models Plugin:** I think here we are very close, it only offers one task to compile models.

- **The Module Plugin:** I agree that especially the publishing part is not ideal, but for people who want to play around with NeonBee it will work out.

So the purpose of this plugin is NOT to simply integrate NeonBee into a complex Gradle project that already exists. It's more focused on people who want to start from scratch and play around and explore NeonBee.

Of course, it makes perfect sense to extract the "base tasks" you mention into a separate Gradle plugin, which can then be used by this plugin or used in a more complex Gradle project that also wants to create "NeonBee modules".

But I think just having the "basic tasks" you mention won't lower the barrier to entry as much as my suggested plugin. Maybe we can agree to release this plugin inside `io.neonbee.gradle.kickstart.*`. This would also make the scope or target audience of this plugin clearer.

pk-work:
https://github.com/SAP/neonbee-plugin-kickstart

Gradle publication is pending.

PR #118, opened 2022-04-11
[POC] feat: add health checks

### LINKED ISSUE #117: Health Checks
## Description

As a NeonBee user, I would like to have a `/health` endpoint, which tells me if NeonBee itself (e.g. the cluster building), and application specific parts (e.g. a db connection) are in a healthy state.

This information should be exposed in an easily consumable way, such as a health endpoint.

## Solution

Hazelcast already provides information about the health of a cluster:

```console
curl -s http://localhost:5701/hazelcast/health | jq -r .
{
  "nodeState": "ACTIVE",
  "clusterState": "ACTIVE",
  "clusterSafe": true,
  "migrationQueueSize": 0,
  "clusterSize": 11
}
```

This information could be used as a starting point for neonbee health checks. With the [vertx-health-check plugin](https://vertx.io/docs/vertx-health-check/java/) we could provide an extensible health endpoint, which - at the beginning - provides a few default checks such as checking the available memory. In case that NeonBee was started in clustered mode, we may want to check the cluster status as well.

Example response of the `/health` endpoint:

```console
root@web-7fc6cbf8f7-jcprt:/# curl -s http://localhost:8080/health | jq -r .
{
  "status": "UP",
  "checks": [
    {
      "id": "physical-memory-check",
      "status": "UP"
    },
    {
      "id": "cluster-health",
      "status": "UP"
    },
    {
      "id": "member-health",
      "status": "UP"
    },
    {
      "id": "database-health",
      "status": "UP",
      "data": {
        "connections": 32
      }
    }
  ],
  "outcome": "UP"
}
```

In order to provide custom (application-specific) health checks, we could use an SPI based approach.

s4heid:
Relates to #117

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     implementation group: 'io.vertx', name: 'vertx-auth-jwt', version: vertx_version
     implementation group: 'io.vertx', name: 'vertx-auth-oauth2', version: vertx_version
     implementation group: 'io.vertx', name: 'vertx-hazelcast', version: vertx_version
+    implementation group: 'io.vertx', name: 'vertx-health-check', version: vertx_version
 
     def jackson_version = '2.13.1'
     implementation group: 'com.fasterxml.jackson.dataformat', name: 'jackson-dataformat-yaml', version: jackson_version
--- a/docs/health-checks.md
+++ b/docs/health-checks.md
@@ -1,0 +1,67 @@
+# Health Checks
+
+NeonBee provides a customizable health check registry, which can be extended
+with application-specific health checks. The results of the health check
+procedures are published via an health endpoint.
+
+## Types of health checks
+
+| name      | enabled    | description                                                           |
+|-----------|------------|-----------------------------------------------------------------------|
+| `memory`  | always     | checks if physical memory is below a threshold of 10%                 |
+| `cluster` | clustered  | checks if cluster is in a safe state and lifecycle service is running |
+| `node`    | clustered  | checks if local member is in a safe state                             |
+
+### Custom Health Checks
+
+Custom health checks can be added by implementing the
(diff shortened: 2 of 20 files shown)


pk-work:
Hi @s4heid ,

I really like the idea of this change!!!

Here is the summary from our discussion in the committer meeting:

## Types
- global checks: Global checks only need to be performed once in the cluster, the cluster node that performs the check doesn't matter.

- node-specific checks: Node-specific checks must be performed on each cluster node, since the result of the check is node-specific.

## Naming
- Global checks should be named with this rule <prefix>.<checkName>. Prefix can be anything but "node".

- Node specific checks should be named with this rule <prefix>.<nodeId>.<checkName>. Prefix MUST be "node". This type of "nesting" is required because the Vert.x HealthCheck endpoint only offers a flat array containing all the checks.

## Soe ideas how an API could look like

HealthCheck.createGlobal(String prefix, String name, vertx -> {
// perform check // should return a Future
}) // Future<HealthCheck>

HealthCheck.createNodeSpecific(String name, vertx -> {
// prefix is fix and nodeID should be retrieved directly from NeonBee
// perform check // should return a Future
}) // Future<HealthCheck>

// Instance of HealthCheck
healthCheck.unregister() // void
healthCheck.getId() // String

// HealthCheck Registry
neonBee.getHealthChecks.register(HealthCheck hc) // Void
neonBee.getHealthChecks.uregister(String checkId)
neonBee.getHealthChecks.getAll() // List<HealthCheck>
neonBee.getHealthChecks.get(String checkId) // HealthCheck
neonBee.getHealthChecks.isActive(String checkId) // Boolean

s4heid:
The wrapper is necessary, because the vertx `HealthChecks` does not prevent overriding of already existing procedures.

s4heid:
Will be implemented with #131.

PR #121, opened 2022-04-20
fix: execute LauncherPreProcessor before configureLogging

halber:
This is changed because the LauncherPreProcessor is used, among other
things, to copy the logging configuration.

### DIFF
--- a/src/main/java/io/neonbee/Launcher.java
+++ b/src/main/java/io/neonbee/Launcher.java
@@ -63,14 +63,14 @@ public class Launcher {
         NeonBeeOptions.Mutable options = new NeonBeeOptions.Mutable();
         CLIConfigurator.inject(commandLine, options);
 
-        // do not use a Logger before this!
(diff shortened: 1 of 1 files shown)


ada89 (approved):
lgtm

kristian:
Hey @halber, one idea that came into my mind while thinking about, that you are unsatisfied with this, because if a logger gets initialized inside of the prelaunchhook, it might mess up our configuration. I don't know if it will actually work but:

Our main issue is that the logger initialization is static. We might be able to overcome that by initializing an own class-loader (which will initialize it's own static instances) just for the purpose of initialization and "throwing it away" afterwards. That might be a very good way of "isolation" anyways, because essentially we would prefer not to "pollute" any of the "real" class-loader with stuff that is happening in pre-loading. What do you think?

I don't know if that is actually feasible for logging though.

halber:
This change caused an error and therefore will not be merged and closed.

PR #123, opened 2022-04-27
refactor: load endpoints asynchronous

pk-work:
In order to be able to use Vert.x OpenAPI, it is required that NeonBee
can load its endpoints asynchronous.


s4heid (dismissed):
This is a good refactoring as it reduces the complexity of the bloated ServerVerticle. Couldn't find any issues. 🚀 

As already discussed, it would be good to mention the breaking change in the endpoint API in the commit message.

s4heid (dismissed):

pk-work:
I adjusted the body of the commit msg

s4heid (approved):

PR #124, opened 2022-04-28
feat: add abstract OpenAPI endpoint

pk-work:


s4heid:
@pk-work what happened to our automated checks? Any clue why they didn't run in this PR?

src/test/java/io/neonbee/endpoint/openapi/AbstractOpenAPIEndpointTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    @DisplayName("createEndpointRouter should fail")
+    void testCreateEndpointRouterFailed(Vertx vertx, VertxTestContext testContext) {
+        Checkpoint methodCheckpoint = testContext.checkpoint();
+        Checkpoint failingCheckpoint = testContext.checkpoint();
+        DummyOpenAPIEndpoint dummyEndpoint = new DummyOpenAPIEndpoint() {
+            @Override
+            protected Future<String> getOpenAPIContractURL(Vertx vertx, JsonObject config) {
+                methodCheckpoint.flag();
+                return succeededFuture("Invalid");
```
s4heid, line 60: This only covers the error case that a contract url was found, but it is invalid. There is also the case that a failed future is returned and we should check that we propagate the error correctly.

s4heid (changes requested):

PR #125, opened 2022-04-29
fix: enable tests to run in intellij idea with coverage

pk-work:
IntelliJ IDEA seems to inject a bunch of libraries to the
classpath/classloader when running tests in coverage mode. This leads to
issues when running tests where classpath scanning is involved, because
for example classes like:
org.jetbrains.coverage.org.objectweb.asm.ClassReader and
org.jetbrains.coverage.org.objectweb.asm.Opcodes (as well as many more)
are found during the classpath scan even if they not match the lookup
criteria.

Therefore, the list of found files is filtered to clean it up from
IntelliJ IDEA injected classes.

Co-authored-by: @gedack


kristian (approved):
lgtm

PR #128, opened 2022-05-12
fix: NeonBee config not loaded

halber:
Tried to load the NeonBee configuration from the wrong place.


s4heid (commented):
The `startNeonBee` method looks like it could be easily tested. Seems like this is currently missing. Would be nice if we could add an assertion that we are passing the correct directory to the `load` method in order to prevent such issues in the future.

halber:
Unit test added.

s4heid (approved):

PR #131, opened 2022-05-23
feat: add health checks

### LINKED ISSUE #117: Health Checks

s4heid:
In order to provide a generic and extendable way of creating health
checks in NeonBee, a central registry is needed where health checks can
be provided. To achieve this, a `HealthCheckRegistry` is added to
NeonBee which will keep track of all default health checks of NeonBee
and custom health checks provided by the NeonBee user.

Vert.x already comes with a health check plugin which will be used for
the health check objects that can be registered on to the registry.
However, as the plugin does not prevent overriding of already existing
health check procedures, a wrapper is necessary to provide this
functionality. In addition, a caching mechanism is implemented which can
be controlled via the retention time. The retention time is the time
from the last health check until a new health check procedure is
triggered. If the next health check is inside this retention window, the
cached result of the previous health check will be returned.

The health checks can be be globally enabled / disabled and a timeout
for the checks can be set via the NeonBeeConfig.

Resolves #117

### DIFF
--- a/src/main/java/io/neonbee/health/HealthCheckRegistry.java
+++ b/src/main/java/io/neonbee/health/HealthCheckRegistry.java
@@ -1,0 +1,176 @@
+package io.neonbee.health;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.neonbee.NeonBee;
+import io.neonbee.health.internal.HealthCheck;
+import io.neonbee.logging.LoggingFacade;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.healthchecks.CheckResult;
+import io.vertx.ext.healthchecks.HealthChecks;
+import io.vertx.ext.healthchecks.Status;
+
+public class HealthCheckRegistry {
+    private static final LoggingFacade LOGGER = LoggingFacade.create();
+
+    @VisibleForTesting
+    HealthChecks healthChecks;
+
+    @VisibleForTesting
+    final Map<String, HealthCheck> checks;
+
+    private final Map<String, HealthCheck> immutableChecks;
+
+    private final Vertx vertx;
+
+    /**
+     * Constructs a new instance of {@link HealthCheckRegistry}.
+     *
+     * @param vertx the current Vert.x instance
+     */
+    public HealthCheckRegistry(Vertx vertx) {
+        this.vertx = vertx;
+        checks = new HashMap<>();
+        immutableChecks = Collections.unmodifiableMap(checks);
+        healthChecks = HealthChecks.create(vertx);
+    }
+
+    /**
+     * Get a map with all health checks registered on this node, with the id of the health check as key.
+     *
+     * @return the health checks map
+     */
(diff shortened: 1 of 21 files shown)


src/main/java/io/neonbee/health/internal/HealthCheck.java
```diff
@@ -0,0 +1,39 @@
...
+     * checks, in which a cached result will be returned before a new health check will be requested again.
+     *
+     * @return the time (in seconds) how long the health check result should be returned from the cache.
+     */
+    long getRetentionTime();
+
+    /**
+     * Indicates whether the health check is a globally scoped health check, or a node-specific health check.
+     * Node-specific health checks perform a check on every single node of a cluster, which is redundant when NeonBee
+     * was started non-clustered.
```
pk-work, line 26:
```suggestion
     * A node specific health check should be used for parameters that can be different on every node in a cluster like RAM or CPU.
```

src/main/java/io/neonbee/health/HealthCheckRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Registers a node health check, which checks all nodes in the cluster. If NeonBee was started non-clustered, the
+     * health check will be applied to the single node.
+     *
+     * @param id            the id of the health check
+     * @param retentionTime the duration how long a cached health status should be returned
+     * @param procedure     the health check procedure that should be registered
+     * @param config        the health check config
+     * @return the registered health check
```
pk-work, line 78:
```suggestion
     * @return the registered health check, or null if health checks are disabled globally
```

pk-work (changes requested):

src/main/java/io/neonbee/health/HealthCheckRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+        healthChecks = HealthChecks.create(vertx);
+    }
+
+    /**
+     * Get a map with all health checks, with the id of the health check as key.
+     *
+     * @return the health checks map
+     */
+    public Map<String, HealthCheck> getHealthChecks() {
+        return this.checks;
```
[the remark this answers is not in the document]
s4heid: Agreed. checks could be misused without the unmodifiable list.

pk-work (approved):

PR #132, opened 2022-05-25
feat: ensure compatibility with Java 17

pk-work:


.github/workflows/voter.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        java: ['11.0']
+        java: ['17.0']
```
s4heid, line 20: why not testing both? would be good to support 11, too. at least for a while.

```suggestion
        java: ['11.0', '17.0']
```

s4heid (changes requested):

s4heid (dismissed):

kristian:
We must continue to support and test both, Java 11 and Java 17. We are not ready to switch to Java 17 for the time being, thus, if we cannot test both, we cannot merge this change.

kristian (changes requested):
We must continue to support Java 11 for the time being

s4heid:
I discussed with Pascal that we would first release a new neonbee version and then merge the change so we have enough time to migrate to 17. @kristian  you have concerns why we can't upgrade maybe we should discuss this in the upcoming committer meeting.

kristian:
Yes let‘s bring this into next committers meeting please. I don‘t see us in a good position migrating currently. Do we have a particular reason why an upgrade is required? So any feature that we particularly need? Vert.x suggesting they will discontinue Java 11 anytime soon? Otherwise I see the issue that the barrier of entry / stake of developing with NeonBee becomes even higher… I don‘t know what the current adoption of Java 17 is amongst developers, however I think in comparison it is still a minority. What speaks against supporting multiple (and also other? Java 13?!) Java LTRs? Thinking about Java 11, 13 and 17 maybe?

pk-work:
If we want to use new Java features, like sealed classes or multi-line string, we need to increase the `sourceCompatibility`. If we don't want that, we can still stick to Java 11. I will add the tests for Java 11 again.

pk-work:
Seems to work with Java 11. But I will then rephrase commit message, because we don't switch :(

kristian:
> If we want to use new Java features, like sealed classes or multi-line string, we need to increase the `sourceCompatibility`. If we don't want that, we can still stick to Java 11. I will add the tests for Java 11 again.

I am not talking about „us“ specifically… but we have to think about everone who wants to run NeonBee as well… this is what I meant essentially: I don‘t think just that we can use multi-line strings justifies us deprecating Java 11. And sure, if we consider using sealed classes, we should consider using Java 17, but are we considering using sealed classes? A switch „just to have switched“ seems a little senseless to me. Let‘s bring this up in our next committers meeting.

pk-work:
Okay, let's discuss this in the next meeting. Do you feel comfortable with the current PR and do you consider to approve it?

kristian:
In the publish GitHub action, is there any issue with only listing Java 17 there?

pk-work:
My understanding is that 'sourceCompatibility' is the important configuration, and the JVM version which is used to build it doesn't matter, but if you feel more comfortable I can switch it back to 11.

PR #137, opened 2022-06-10
fix(health): succeed if health check config not found

s4heid:
This change ensures that reading the health check config file from the
config directory succeeds in case the file does not exist and returns
the default configuration otherwise.

### DIFF
--- a/src/test/java/io/neonbee/health/AbstractHealthCheckTest.java
+++ b/src/test/java/io/neonbee/health/AbstractHealthCheckTest.java
@@ -3,6 +3,8 @@ package io.neonbee.health;
 import static com.google.common.truth.Truth.assertThat;
 import static io.neonbee.health.AbstractHealthCheck.DEFAULT_RETENTION_TIME;
 import static io.neonbee.test.helper.OptionsHelper.defaultOptions;
+import static io.vertx.core.Future.failedFuture;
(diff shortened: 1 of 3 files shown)


src/test/java/io/neonbee/health/AbstractHealthCheckTest.java
```diff
@@ -79,12 +98,74 @@ void testRegisterNodeCheck(VertxTestContext testContext) throws HealthCheckExcep
...
+        @Override
+        public Stream<? extends Arguments> provideArguments(ExtensionContext context) {
+            return Stream.of(Arguments.of(succeededFuture(new JsonObject().toBuffer()), "config from file is empty"),
+                    Arguments.of(succeededFuture(Buffer.buffer()), "config from file is an empty buffer"),
+                    Arguments.of(failedFuture("oops"), "retrieving config fails"));
+        }
+    }
+
+    @ParameterizedTest(name = "{index} => {1}")
+    @ArgumentsSource(CustomConfigArgumentProvider.class)
```
pk-work, line 135: Why you do this via a provider class instead of a provider method?

s4heid: No specific reason :)

halber (approved):
LGTM

PR #140, opened 2022-06-13
Make health checks addable via SPI

s4heid:
Depends on #136 and #139


pk-work (approved):
Approved, but we should open an issue because this is now the 4th variant of loading classes during runtime. We should open an issue to remove this technical debt and check if we can have one unified way of loading classes at runtime.

PR #142, opened 2022-06-23
feat: support response hint

dapeng-wang:


kristian:
RoutingContext routingContext
...
DataContext context

DataRequest request = new DataRequest("target");
request.putHeader("...", "...")

- <DataResponse<T>> requireData(...)

DataResponse
data() --> <T>
headers() --> Map<String, Object>

- DataContext.class:

Map<String, Object> data();
  mergeData(...) // AUTO_MERGE / MANUAL_MERGE, in den DataVerticle eine protected mergeContextData() { ... }

Map<String, Object> responseData(request)

<-- responseData Map<String, Object>, DataContextImpl Map<DataRequest, Map<String, Object>>

Future<String> retriveData(DataQuery, DataMap, DataContext context) {
	...
	DataRequest request = new DataRequest("target", new DataQuery());
	requestData(request, context).compose(result -> {
		Map<String, Object> globalgalaktischeMap = context.data();

		Map<String, Object> responseData = context.responseData(request);
		context.putResponseData("context-type-hint", responseData.get("context-type-hint");
        });
	...
}

kristian:
As promised I made my mind up about the naming scheme here. As we discussed we have a "response" channel of the verticles that want to send data to the upstream verticle (currently handled via `context.responseData().put()) and then for the upstream / downstream verticle to access this information (currently handled via `context.downstreamResponseData().get()`. One of the main problems here is for me confusing both of them, because depending on the "direction of thinking" in your head downstream / upstream, or inbound / outbound or response / requesst, etc. can always be confused. For example:

Does "responseData" mean the data that "was the response" of the verticles I called or is it the "data that I as a verticle being called will respond with"? (at the moment it is the latter).

And this can be applied to all combinations. Stuff like "downstreamResponseData" is kind of redundant and you also again are in the position of the viewer or the sender here, right: So am I as a the consumer of the verticle really "downstream"? Because I would argue, only in the eye of the receiving verticle that so to speak sets the data, I am downstream... So this makes naming for me very difficult...

And based on our todays discussion @dapeng-wang  I think your proposal is the best:

``` 
context.reponseData().put(...) // to return data to the callee, whilst .get() etc. still work like in a normal map and
context.receivedData().get(...) // to read data after the called returned from teh called verticle, whilst .put() on this map should result in an unsupportedoperation exception (so make the map unmodifable) to also clearly indiciate to the end-user that this is only the "read-side" interface of the both
```

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+     */
+    Map<String, Object> responseData();
+
+    /**
+     * Sets all the response data.
+     *
+     * @param data the data, must be compatible to {@link JsonObject#JsonObject(Map)}
+     * @return a reference to this DataContext for chaining
+     */
+    DataContext setResponseData(Map<String, Object> data);
```
kristian, line 115: not needed, everything that is to be changed on the responseData should be changed on the responseData() map itself, e.g. responseData().putAll(...)

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+
+    /**
+     * Merges the response data of a given map into the existing response context data map.
+     * <p>
+     * Note: This method does not perform a deep merge operation, but overrides already existing elements w/o merging
+     *
+     * @param data the response data to merge, must be compatible to {@link JsonObject#JsonObject(Map)}
+     * @return a reference to this DataContext for chaining
+     */
+    DataContext mergeResponseData(Map<String, Object> data);
```
kristian, line 125: Same

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+     * @return the previous data associated with the key, as expected from {@link JsonObject#remove(String)}
+     */
+    <T> T removeResponseDataEntry(String key);
+
+    /**
+     * Returns a map between {@link DataRequest} and received data map.
+     *
+     * @return a map between {@link DataRequest} and received data map
+     */
+    Map<DataRequest, Map<String, Object>> receivedData();
```
kristian, line 123: have you considered using a DataMap<Object> here?

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+
+    /**
+     * Sets the response data map for all {@link DataRequest}s.
+     *
+     * @param map a map between {@link DataRequest} and received response data for the request.
+     * @return a reference to this DataContext for chaining
+     */
+    DataContext setReceivedData(Map<DataRequest, Map<String, Object>> map);
```
kristian, line 131: this should never be possible to be called by an end-user, so please remove

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+     */
+    Optional<Map<String, Object>> findFirstReceivedData(String qualifiedName);
+
+    /**
+     * Retrieves all received data maps for all verticles identified by the name.
+     *
+     * @param qualifiedName qualified name of a verticle
+     * @return all received data maps from verticles with the name
+     */
+    List<Map<String, Object>> findAllReceivedData(String qualifiedName);
```
kristian, line 155: if receivedData() would return da DataMap, we could get rid of all these methods as well and the interfaces of DataMap and DataContext would stay consistent regardless of what we do.

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,105 @@ public interface DataContext {
...
+
+    /**
+     * Merge the received data from all invoked verticles into the current verticle's context.
+     */
+    void mergeReceivedDataIntoResponseData();
```
kristian, line 160: How about "propagateReceivedData()"?

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -75,11 +75,206 @@
 
     @SuppressWarnings("UnnecessaryLambda") // overridden in DummyVerticleHelper, as getNamespace is final
     private final Supplier<String> namespaceSupplier =
-            () -> Optional.ofNullable(this.getClass().getAnnotation(NeonBeeDeployable.class))
```
kristian, line 78: We decided to not staticially import methods that do not describe what they create, e.g.:

Map.of
List.of
Optional.of

Etc. otherwise there could be ambiguities

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -75,11 +75,206 @@
...
+                break;
+            default:
+                failureCode = replyException.failureCode();
+                break;
+            }
+        }
+
+        return new DataException(failureCode, message);
+    }
+
```
kristian, line 277: why was all of this moved? this makesreviewing next to impossible, can you move it back and move it in another change if required?

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -104,6 +299,18 @@
...
+     * verticle to handle the merging and determine the returned metadata. If AUTO_MERGE is returned, the metadata of
+     * all invoked verticles will be automatically merged and returned. The default is set to MANUAL_PROCESS, each
+     * verticle can overwrite this method to use a different policy.
+     *
+     * @return how this verticle handles the response metadata.
+     */
+    public ResponseDataPropagationPolicy getResponseDataPropagationPolicy() {
+        return ResponseDataPropagationPolicy.MANUAL_PROCESS;
+    }
+
```
kristian, line 313: As we now have a context.propagateresponseData() method, do we really need this? Seem like overkill to me: So in order to do the propagation, I would have to:

```java
@Override
public ResponseDataPropagationPolicy getResponseDataPropagationPolicy() {
        return ResponseDataPropagationPolicy.AUTO_MERGE;
}
```

And this causes some "magic" to happen. Instead I could simply (in my already implemented `retrieveData` call:

```java
@Override
public Future<?> retrieveData() {
        // ...
        context.propagateResponseData();
        // ...
}
```

Before returning the method. Doesn't the second one fully cover the whole functionallity much more easily and more clear to the end user, because he can influence when and where he wants to do the propagation?

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -183,10 +390,11 @@ public void start(Promise<Void> promise) {
             ResolutionRoutine routine;
             MultiMap headers = message.headers();
             try {
-                routine = message.body().getAction() == READ
-                        ? resolutionRoutineForStrategy(Optional.ofNullable(headers.get(RESOLUTION_STRATEGY_HEADER))
-                                .map(ResolutionStrategy::valueOf).orElse(RECURSIVE))
-                        : new ManipulationRoutine();
+                routine =
+                        message.body().getAction() == READ
+                                ? resolutionRoutineForStrategy(ofNullable(headers.get(RESOLUTION_STRATEGY_HEADER))
```
kristian, line 395: Same here no default import of "of" methods please. (We should add that to our coding conventions in the docs folder please)

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -398,120 +529,6 @@ public final String getQualifiedName() {
...
-                break;
-            default:
-                failureCode = replyException.failureCode();
-                break;
-            }
-        }
-
-        return new DataException(failureCode, message);
-    }
-
```
kristian, line 514: all these lines have been moved around, review is impossible

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -528,7 +545,7 @@ private ResolutionRoutine resolutionRoutineForStrategy(ResolutionStrategy strate
 
     /**
      * Interface for all resolution routines (actual implementations of resolution strategies).
-     *
+     * <p>
```
kristian, line 548: ? unnecesarry

kristian (changes requested):
Ich bin nicht durch gekommen

src/main/java/io/neonbee/data/DataContext.java
dapeng-wang: I don't think DataMap is a good fit, since we are dealing with key value pairs.

It is unnessary to wrap the key value pairs into an AsyncResult.

src/main/java/io/neonbee/data/DataContext.java
dapeng-wang: A setter is but required to set the received data in DataVerticle (RecursiveResolutionRoutine) into the DataContext.

Since DataContextImpl is not in the same package as DataVerticle, a package protected method will not work.

src/main/java/io/neonbee/data/DataContext.java
dapeng-wang: See above.

src/main/java/io/neonbee/data/DataVerticle.java
dapeng-wang: Done

src/main/java/io/neonbee/data/DataVerticle.java
dapeng-wang: The root cause is that some rule was introduced to enforced the order of the public protected and private methods. I didn't check exactly whether in PMD oder Sonar.

Without moving the code, my PR build always failed.

I don't like it either.

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,59 @@ public interface DataContext {
...
+
+    /**
+     * Merges the response data of a given map into the existing response context data map.
+     * <p>
+     * Note: This method does not perform a deep merge operation, but overrides already existing elements w/o merging
+     *
+     * @param data the response data to merge, must be compatible to {@link JsonObject#JsonObject(Map)}
+     * @return a reference to this DataContext for chaining
+     */
+    DataContext mergeResponseData(Map<String, Object> data);
```
kristian, line 116: Same here? Better fit into DataContextImpl, correct? No user should call this, this is only to be called by the framework

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,59 @@ public interface DataContext {
...
+    /**
+     * Retrieves all received data maps for all verticles identified by the name.
+     *
+     * @param qualifiedName qualified name of a verticle
+     * @return all received data maps from verticles with the name
+     */
+    List<Map<String, Object>> findAllReceivedData(String qualifiedName);
+
+    /**
+     * Propagate the received data from all invoked verticles into the current verticle's context.
```
kristian, line 150: Maybe just a little text suggestion here:

Propagate all received data intothe current verticle's response data.

@see #responseData

kristian (changes requested):

src/main/java/io/neonbee/data/DataContext.java
```diff
@@ -96,6 +98,49 @@ public interface DataContext {
...
+     * Arbitrary response data of this context, which is passed backwards from callee to caller.
+     *
+     * @return all the context response data as a map.
+     */
+    Map<String, Object> responseData();
+
+    /**
+     * Returns a map between {@link DataRequest} and received data map.
+     *
+     * @return a map between {@link DataRequest} and received data map
```
kristian, line 111: this JavaDoc could be better explaining how responseData / receivedData is to be used and also that some helper methods are available, e.g. findReceivedData, etc.

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -343,10 +343,14 @@ public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataCo
...
 
                         if (asyncReply.succeeded()) {
+                            DataContext responseDataContext =
+                                    decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER));
                             context.setData(Optional
                                     .ofNullable(
                                             decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER)))
                                     .map(DataContext::data).orElse(null));
+                            ((DataContextImpl) context).mergeResponseData(Optional.ofNullable(responseDataContext)
+                                    .map(DataContext::responseData).orElse(null));
```
kristian, line 353: I don't entirely understand this one? as soon as data is received back, we always merge the responseData we received into our own responseData? isn't it that we have to put this into our receivedData instead?! only if propagation is done via the propageteResponseData method then this may be called, isn't it?

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -343,10 +343,14 @@ public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataCo
                         LOGGER.correlateWith(context).debug("Received event bus reply");
 
                         if (asyncReply.succeeded()) {
+                            DataContext responseDataContext =
+                                    decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER));
                             context.setData(Optional
                                     .ofNullable(
                                             decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER)))
                                     .map(DataContext::data).orElse(null));
```
kristian, line 351: decodeFromString should not be called twice here, because it is kind of a heavy operation, but maybe it is not needed at all, see comment below

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -548,18 +552,28 @@ public Future<T> execute(DataQuery query, DataContext context) {
...
-                        .map(request -> requestResults.computeIfAbsent(request, mapRequest -> {
-                            Future<Object> future = requestData(vertx, request, context.copy());
-                            reportRequestDataMetrics(request, future);
-                            return future;
-                        })).map(Future.class::cast).collect(Collectors.toList())).otherwiseEmpty();
+                return CompositeFuture.join(
+                        Optional.ofNullable(requests).map(Collection::stream).orElse(Stream.empty()).map(request -> {
+                            // use one copy of DataContext for each request to avoid data clash
+                            DataContext requestContext = context.copy();
+                            receivedDataContextMap.put(request, requestContext);
```
kristian, line 563: does it make sense to only already store the receivedData here, then we could maybe spare us the stream().map operation of the map in line 573-575 or why else is the whole map of the context required?

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -548,18 +552,28 @@ public Future<T> execute(DataQuery query, DataContext context) {
...
+                                reportRequestDataMetrics(request, future);
+                                return future;
+                            });
+                        }).map(Future.class::cast).collect(Collectors.toList())).otherwiseEmpty();
             }).compose(requiredCompositeOrNothing -> {
                 List<Tag> tags = retrieveDataTags();
                 try {
+                    Map<DataRequest, Map<String, Object>> receivedData = receivedDataContextMap.entrySet().stream()
+                            .map(entry -> Map.entry(entry.getKey(), entry.getValue().responseData()))
+                            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
```
kristian, line 575: Could may be be spared out, see comment above

src/main/java/io/neonbee/data/internal/DataContextImpl.java
```diff
@@ -52,6 +54,8 @@ public class DataContextImpl implements DataContext {
 
     private static final Pattern BEARER_AUTHENTICATION_PATTERN = Pattern.compile("Bearer\\s(.+)");
 
+    private static final String RESPONSE_METADATA_KEY = "responsedata";
+
```
kristian, line 58: you could put this one above to the other KEY constants also here it still says Metadata, so suggestion would be:

private static final String RESPONSE_DATA_KEY = "responseData";

src/main/java/io/neonbee/data/internal/DataContextImpl.java
```diff
@@ -104,16 +116,24 @@ public DataContextImpl(String correlationId, String bearerToken, JsonObject user
...
+    public DataContextImpl(String correlationId, String sessionId, String bearerToken, JsonObject userPrincipal,
+            Map<String, Object> data, Map<String, Object> resposneData, Deque<DataVerticleCoordinate> paths) {
         this.correlationId = correlationId;
         this.sessionId = sessionId;
         this.bearerToken = bearerToken;
         // actually create a read only user principal object, so that no one can tamper with the data
         this.userPrincipal = Optional.ofNullable(userPrincipal).map(JsonObject::getMap)
                 .map(Collections::unmodifiableMap).map(JsonObject::new).orElse(null);
         this.setData(data); // create a mutable copy of the map
+        this.responseData = resposneData;
```
kristian, line 136: typo and the resposneData that is passed in might not be mutable, so similar to setData better create a mutable copy?

src/main/java/io/neonbee/data/internal/DataContextImpl.java
```diff
@@ -241,7 +325,8 @@ public static String encodeContextToString(DataContext context) {
         }
         return new JsonObject().put(CORRELATION_ID, context.correlationId()).put(SESSION_ID_KEY, context.sessionId())
                 .put(BEARER_TOKEN_KEY, context.bearerToken()).put(USER_PRINCIPAL_KEY, context.userPrincipal())
-                .put(DATA_KEY, context.data()).put(PATH_KEY, pathToJson(context.path())).toString();
+                .put(DATA_KEY, context.data()).put(RESPONSE_METADATA_KEY, context.responseData())
```
kristian, line 328: does this work? as we have a map of Map<String, Object> we can use the new JsonObject(Map) constructor here... I don't think making a put that is a map is supported by JsonObject is it?

also same comment as for adela in the other change, see comment in the `encDetail` line at: https://github.com/SAP/neonbee/pull/165#discussion_r955231128

I think we should make it clear to the user int he JavaDoc comment of responseData(), that the only supported values of the Map<String, Object> are values which are supported by the JsonObject(Map) constructor in cluste roperation mode.

kristian (changes requested):
Very good change DaPeng! Thanks for all the work put into this and sorry for the "concept phase" that held you up :) we try to improve that moving forward!

Just some smaller things and change lgtm after that! 👍 thanks again!

src/main/java/io/neonbee/data/DataVerticle.java
dapeng-wang: I don't think so, we are keeping the reference to the DataContext before sending the requests via the CompositeFuture. At this time, we don't have reference to the receivedData, since they are not yet allocated at this time. The only reference we have is the DataContext wrapper.

src/main/java/io/neonbee/data/internal/DataContextImpl.java
dapeng-wang: We have been using this code put(DATA_KEY, context.data()) since years, which is also a Map<String, Object>.

Maybe we never exchanged other types than basic types and String.

Nevertheless, I added the map to Jsonobject conversion.

src/main/java/io/neonbee/data/DataVerticle.java
dapeng-wang: the data context will be encoded and decoded when tranferred via event-bus. Here we are receiving the decocded data with updates from eventbus.

src/main/java/io/neonbee/data/internal/DataContextImpl.java
dapeng-wang: Done

s4heid:
@dapeng-wang please squash these commits together and we can merge :) thank you

PR #143, opened 2022-06-28
fix: vert.x worker threads included in StaleThreadChecker

halber:
The StaleThreadChecker currently checks for all vert.x threads. This also includes the worker threads. However, a worker thread may be blocked and should therefore not be included in this check.


pk-work (dismissed):

kristian:
If worker threads are blocked after a test has ended, there is something wrong with the test and schould be reported by the StaleThreadChecker, or do I understand it wrong?

halber:
> If worker threads are blocked after a test has ended, there is something wrong with the test and schould be reported by the StaleThreadChecker, or do I understand it wrong?
I completely agree on that. It looks like I need to take another look at the problem I'm facing in a unit test.

PR #144, opened 2022-06-30
feat: Make ServerVerticle handler configurable.

halber:
The `ServerVerticle` attaches request handlers to the route. The request handlers attached to the route cannot be changed at the moment. This change makes the request handlers that the `ServerVerticle` attaches to the route configurable.

A `RoutingHandlerFactory` is used to create an instance of the `Handler<RoutingContext>` class. In the `ServerConfig`, a list of `RoutingHandlerFactory` class names is used to specify the `RoutingHandlerFactory`. The handler created by the `RoutingHandlerFactory` is then added to the root route. The request handlers are added in the order specified in the `ServerConfig` list.

### DIFF
--- a/docs/ServerVerticle.md
+++ b/docs/ServerVerticle.md
@@ -1,0 +1,25 @@
+# ServerVerticle configuration
+
+## Configuration
+
+### Handler factories
+
+Handler factories are used to register routing handlers on the root router in the ServerVerticle. A handler factory must
+implement the `io.neonbee.internal.handler.factories.RoutingHandlerFactory` interface and must be configured in the
+handlerFactories section of the ServerVerticle configuration. The configured handlerFactories are loaded and added in
+the configured order.
+The order of handler factories has to take the priority of the returned handlers into account.
+See `io.vertx.ext.web.impl.RouteState.weight` If the order is violated, an Exception is thrown.
+
+#### Default configuration
+
+```yaml
+handlerFactories:
+- io.neonbee.internal.handler.factories.LoggerHandlerFactory
+- io.neonbee.internal.handler.factories.InstanceInfoHandlerFactory
+- io.neonbee.internal.handler.factories.CorrelationIdHandlerFactory
+- io.neonbee.internal.handler.factories.TimeoutHandlerFactory
+- io.neonbee.internal.handler.factories.SessionHandlerFactory
+- io.neonbee.internal.handler.factories.CacheControlHandlerFactory
+- io.neonbee.internal.handler.factories.DisallowingFileUploadBodyHandlerFactory
+```
--- a/src/main/java/io/neonbee/internal/handler/factories/SessionHandlerFactory.java
+++ b/src/main/java/io/neonbee/internal/handler/factories/SessionHandlerFactory.java
@@ -1,0 +1,69 @@
+package io.neonbee.internal.handler.factories;
+
+import java.util.Optional;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import io.neonbee.NeonBee;
+import io.neonbee.config.ServerConfig;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.handler.PlatformHandler;
+import io.vertx.ext.web.handler.SessionHandler;
+import io.vertx.ext.web.sstore.ClusteredSessionStore;
+import io.vertx.ext.web.sstore.LocalSessionStore;
+import io.vertx.ext.web.sstore.SessionStore;
+
+/**
+ * Creates a SessionHandler.
+ *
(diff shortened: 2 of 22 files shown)


BenniR6 (commented):
-No singletons needed for handlers as these are only used on start up.
-Setting the allow file upload to true is missing in this PR -> as already mentioned

**Documentation will follow.**

halber:
> -Setting the allow file upload to true is missing in this PR -> as already mentioned
Maybe we have misunderstood each other. I will not add such a handler because at least in my option such ha handler is not required and can't be implemented easily because the implementation depends heavily on the thing needed.

halber:
Do not merge! there is an issue I currently not understand.

pk-work:
Concept was validated in the committers meeting - 2022-08-02

src/main/java/io/neonbee/internal/handler/factories/SessionHandlerFactory.java
```diff
@@ -0,0 +1,68 @@
...
+import io.vertx.ext.web.handler.PlatformHandler;
+import io.vertx.ext.web.handler.SessionHandler;
+import io.vertx.ext.web.sstore.ClusteredSessionStore;
+import io.vertx.ext.web.sstore.LocalSessionStore;
+import io.vertx.ext.web.sstore.SessionStore;
+
+/**
+ * Creates a SessionHandler.
+ *
+ * If ServerConfig.SessionHandling is set to NONE a no operation handler is returned;
```
s4heid, line 22:
```suggestion
 * If ServerConfig.SessionHandling is set to NONE a no-operation handler is returned, which does not perform any session handling.
```

src/main/java/io/neonbee/internal/handler/factories/SessionHandlerFactory.java
```diff
@@ -0,0 +1,68 @@
...
+        return Future.succeededFuture(sh);
+    }
+
+    /**
+     * Creates a {@linkplain SessionStore} based on the given {@linkplain ServerConfig} to use either local or clustered
+     * session handling. If no session handling should be used, an empty optional is returned.
+     *
+     * @param vertx           the Vert.x instance to create the {@linkplain SessionStore} for
+     * @param sessionHandling the session handling type
+     * @return a optional session store, suitable for the given Vert.x instance and based on the provided config value
```
s4heid, line 50:
```suggestion
     * @return an optional session store, suitable for the given Vert.x instance and based on the provided config value
```

src/main/java/io/neonbee/internal/handler/factories/SessionHandlerFactory.java
```diff
@@ -0,0 +1,68 @@
...
+     *         (none/local/clustered). In case the session handling is set to clustered, but Vert.x does not run in
+     *         clustered mode, fallback to the local session handling.
+     */
+    @VisibleForTesting
+    static Optional<SessionStore> createSessionStore(Vertx vertx, ServerConfig.SessionHandling sessionHandling) {
+        switch (sessionHandling) {
+        case LOCAL: // sessions are stored locally in a shared local map and only available on this instance
```
s4heid, line 57: Maybe we can move all these code comments into the javadoc where it belongs?

src/test/java/io/neonbee/internal/handler/factories/CorrelationIdHandlerFactoryTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+@ExtendWith(VertxExtension.class)
+class CorrelationIdHandlerFactoryTest {
+
+    @Test
+    void testCreateHandler(VertxTestContext testContext) {
+
+        try (MockedStatic<NeonBee> staticNeonBeeMock = mockStatic(NeonBee.class)) {
+            NeonBee neonBeeMock = mock(NeonBee.class);
+            when(neonBeeMock.getServerConfig()).thenReturn(new ServerConfig());
```
s4heid, line 26: I think you could also also work here with the NeonBeeMockHelper in order to avoid the usage of mockStatic. Not required to change this, but generally I think it's more convenient to use the helper.

s4heid (changes requested):

docs/ServerVerticle.md
```diff
@@ -0,0 +1,24 @@
...
+
+## configuration
+
+### handler factories
+
+Handler factories are used to register routing handlers on the root router in the ServerVerticle. A handler factory must
+implement the `io.neonbee.internal.handler.factories.RoutingHandlerFactory` interface and must be configured in the
+handlerFactories section of the ServerVerticle configuration. The configured handlerFactories are loaded and added in
+the configured order.
+The order of handler factories has to take the priority of the returned handlers into account.
```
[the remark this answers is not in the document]
halber: Unfortunately, I can't find any documentation from vertx for this.

s4heid (approved):

PR #155, opened 2022-07-15
add: TypedDataVerticle RFC 1

eldefay:
add: RFC for Typed DataVerticle

### DIFF
--- a/docs/rfc/NeonBee TypedDataVerticle RFC 1
+++ b/docs/rfc/NeonBee TypedDataVerticle RFC 1
@@ -1,0 +1,90 @@
+# NeonBee TypedDataVerticle RFC 1
+
+### Contents
+- [Abstract](#abstract)
+- [Motivation](#Motivation)
+- [Technical Details](#technical-details)
+- [Performance Considerations](#performance-considerations)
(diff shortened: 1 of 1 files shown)


kristian:
Thanks @eldefay for your first contribution. As discussed last week, I think having a more validated way to exchange JSON data between `DataVerticles` (maybe in form of a `SchemaVerticle`?) and towards an endpoint via JSONSchema would be a great contribution. Let's start the discussion in our next weeks committers meeting. Thanks.

kristian:
Discussion points in NeonBee committers meeting:

Concept is generally accepted, we should drive this further.

Discussion from the meeting:

Could tie in neatly with the OpenAPI endpoint (Pascal will provide a demo)
Maybe possible to "push down" the schema checking from the OpenAPI endpoint to the TypedDataVerticle
Whenever data is requested from eihter soruce (endpoint, other data or entity verticles) from a TypeedDataVerticle the schema is checked

JSON Schema support JsonObject, JsonArray, however in Vert.x there is no common super type for both
DataVerticle<JsonObject> / DataVerticle<JsonArray>
TypedDataVerticle extends DataVerticle<Object> (always returning a JsonObject or Array)
requireData<TypedDataVerticle>(...).componse(object -> ...)

PR #161, opened 2022-08-04
fix: same NeonBee nodeId for multiple instances

halber:
The random NeonBee NODE_ID is the same for all NeonBee instances loaded
from the same classloader. This fix makes the NeonBee nodeId an instance
variable so that each NeonBee instance has its own nodeId.

### DIFF
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -95,7 +95,7 @@ import io.vertx.spi.cluster.hazelcast.HazelcastClusterManager;
 
 @SuppressWarnings({ "PMD.CouplingBetweenObjects", "PMD.GodClass" })
 public class NeonBee {
-    /** // @formatter:off
+    /* // @formatter:off
      *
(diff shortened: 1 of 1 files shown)


s4heid:
@halber I think this shouldn't be an issue, only if we would deploy multiple NeonBee instances on the same node... which should never be the case, right? Has this caused any problems in a scenario I am overlooking?

halber:
> @halber I think this shouldn't be an issue, only if we would deploy multiple NeonBee instances on the same node... which should never be the case, right? Has this caused any problems in a scenario I am overlooking?

Right, it is unlikely that anyone would do this except for tests like in [io.neonbee.NeonBeeExtensionBasedTest#testNeonBeeWithClusters.](https://github.com/SAP/neonbee/blob/7d27776665170b17c338f073cd616ba39ef762bf/src/test/java/io/neonbee/NeonBeeExtensionBasedTest.java#L81) In this case, you can't distinguish the instances by their IDs.

s4heid (approved):

PR #162, opened 2022-08-04
feat: unregister entities from shared map

halber:
NeonBee uses the shared map to manage the registration fo entities to
EntityVerticles e.g:
("Products" -> "ProductsVerticle")
("OrderTypes" -> "C4SOrderTypesVerticle")
During the boot up of an EntityVerticle "announceEntityVerticle" is used to register:
"Entity Products" is handled by "me" (EntityProductsVerticle)
Currently no "stop()" method is implemented, and no fallback handling was implemented (so the map may contain entries of verticles that no longer exist)

This commit adds two methods to delete the registered entities from the shared map.
The `UnregisterEntityModelHook#unregisterOnShutdown` method is executed if a node shuts down gracefully.
The `UnregisterEntityModelHook#cleanup` method is executed when the cluster detects that a node has left

### DIFF
--- a/src/main/java/io/neonbee/internal/cluster/ClusterHelper.java
+++ b/src/main/java/io/neonbee/internal/cluster/ClusterHelper.java
@@ -1,0 +1,106 @@
+package io.neonbee.internal.cluster;
+
+import java.util.Optional;
+import java.util.Set;
+
+import org.infinispan.manager.EmbeddedCacheManager;
+
+import com.hazelcast.cluster.Member;
+
+import io.vertx.core.Vertx;
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.spi.cluster.ClusterManager;
+import io.vertx.ext.cluster.infinispan.InfinispanClusterManager;
+import io.vertx.spi.cluster.hazelcast.HazelcastClusterManager;
+
+public class ClusterHelper {
+
+    /**
+     * Returns the {@link ClusterManager} if NeonBee is started in clustering mode, otherwise it returns null.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return null or the {@link HazelcastClusterManager} instance
+     */
+    public static Optional<ClusterManager> getClusterManager(Vertx vertx) {
+        if (vertx instanceof VertxInternal) {
+            VertxInternal vertxInternal = (VertxInternal) vertx;
+            return Optional.ofNullable(vertxInternal.getClusterManager());
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Returns an optinal with {@link HazelcastClusterManager} if NeonBee was started in cluster mode and
+     * {@link HazelcastClusterManager} is used as {@link ClusterManager}, otherwise it returns an empty Optional.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return null or the {@link HazelcastClusterManager} instance
+     */
+    public static Optional<HazelcastClusterManager> getHazelcastClusterManager(Vertx vertx) {
+        return getSpecificClusterManager(vertx, HazelcastClusterManager.class);
+    }
+
+    /**
+     * Returns an optinal with {@link InfinispanClusterManager} if NeonBee was started in cluster mode and
+     * {@link InfinispanClusterManager} is used as {@link ClusterManager}, otherwise it returns an empty Optional.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return null or the {@link InfinispanClusterManager} instance
+     */
+    public static Optional<InfinispanClusterManager> getInfinispanClusterManager(Vertx vertx) {
+        return getSpecificClusterManager(vertx, InfinispanClusterManager.class);
+    }
+
+    /**
+     * Returns an optinal with the provided {@link ClusterManager} class if NeonBee was started in cluster mode and the
+     * {@link ClusterManager} is used as {@link ClusterManager}, otherwise it returns an empty Optional.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return null or the {@link InfinispanClusterManager} instance
+     */
+    private static <T extends ClusterManager> Optional<T> getSpecificClusterManager(Vertx vertx, Class<T> cmClass) {
+        return getClusterManager(vertx).map(cm -> cmClass.isInstance(cm) ? cmClass.cast(cm) : null);
+    }
+
+    /**
+     * Get the cluster node ID.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return the cluster node ID
+     */
+    public static String getClusterNodeId(Vertx vertx) {
+        return getClusterManager(vertx).map(ClusterManager::getNodeId).orElseThrow(() -> new IllegalStateException(
+                "Can not retrieve the ClusterManager. Is vert.x running in a cluster?"));
+    }
+
+    /**
+     * Is this cluster node the leader?
+     *
+     * @param vertx {@link Vertx} instance
+     * @return true if this node is the leader, otherwise false
+     */
+    public static boolean isLeader(Vertx vertx) {
+        if (!vertx.isClustered()) {
+            return true;
+        }
+
+        return getHazelcastClusterManager(vertx).map(hcm -> isLeader(hcm))
+                .or(() -> getInfinispanClusterManager(vertx).map(icm -> isLeader(icm)))
+                .orElseThrow(() -> new IllegalStateException(
+                        "Can not find the cluster leader. Is vert.x running in a cluster?"));
+    }
+
+    private static Boolean isLeader(InfinispanClusterManager icm) {
+        EmbeddedCacheManager cacheContainer = (EmbeddedCacheManager) icm.getCacheContainer();
+        return cacheContainer.isCoordinator();
+    }
+
+    private static boolean isLeader(HazelcastClusterManager hcm) {
+        // getMembers() returns a set. The first element is the leader.
+        // The underlying set is a kind of linked hashset and In this case the order is kept.
+        // see https://github.com/hazelcast/hazelcast/issues/3760
+        Set<Member> members = hcm.getHazelcastInstance().getCluster().getMembers();
+        Member oldestMember = members.iterator().next();
+        return oldestMember.localMember();
+    }
+}
--- a/src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
+++ b/src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
@@ -1,0 +1,92 @@
+package io.neonbee.internal.cluster.entity;
+
+import static io.neonbee.hook.HookType.CLUSTER_NODE_ID;
+import static io.vertx.core.Future.succeededFuture;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.neonbee.NeonBee;
+import io.neonbee.hook.Hook;
+import io.neonbee.hook.HookContext;
+import io.neonbee.hook.HookType;
+import io.neonbee.internal.Registry;
+import io.neonbee.internal.cluster.ClusterHelper;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+
+/**
+ * Hooks for unregistering verticle models.
+ */
+public class UnregisterEntityVerticlesHook {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UnregisterEntityVerticlesHook.class);
+
+    /**
+     * This method is called when a NeonBee instance shutdown gracefully.
+     *
+     * @param neonBee     the {@link NeonBee} instance
+     * @param hookContext the {@link HookContext}
+     * @param promise     {@link Promise} to complete the function.
+     */
+    @Hook(HookType.BEFORE_SHUTDOWN)
+    public void unregisterOnShutdown(NeonBee neonBee, HookContext hookContext, Promise<Void> promise) {
+        LOGGER.info("Unregistering models on shutdown");
+        Vertx vertx = neonBee.getVertx();
+        String clusterNodeId = ClusterHelper.getClusterNodeId(vertx);
+        unregister(neonBee, clusterNodeId).onComplete(promise)
+                .onSuccess(unused -> LOGGER.info("Models unregistered successfully"))
+                .onFailure(ignoredCause -> LOGGER.error("Failed to unregister models on shutdown"));
+    }
+
+    /**
+     * Unregister the entity qualified names for the node by ID.
+     *
+     * @param neonBee       the {@link NeonBee} instance
+     * @param clusterNodeId the ID of the cluster node
+     * @return Future
+     */
+    public static Future<Void> unregister(NeonBee neonBee, String clusterNodeId) {
+        if (!neonBee.getVertx().isClustered()) {
+            return succeededFuture();
+        }
+
+        Registry<String> registry = neonBee.getEntityRegistry();
+        if (!(registry instanceof ClusterEntityRegistry)) {
+            LOGGER.warn("Running in clustered mode but not using the ClusterEntityRegistry.");
+            return succeededFuture();
+        }
+
+        ClusterEntityRegistry clusterEntityRegistry = (ClusterEntityRegistry) registry;
+
+        LOGGER.info("Unregistering entity verticle models for node ID {} ...", clusterNodeId);
+        Future<Void> unregisterFuture = clusterEntityRegistry.unregisterNode(clusterNodeId);
+
+        return unregisterFuture
+                .onSuccess(
+                        unused -> LOGGER.info("Unregistered entity verticle models for node ID {} ...", clusterNodeId))
+                .onFailure(cause -> LOGGER.error("Failed to unregistered entity verticle models for node ID {} ...",
+                        clusterNodeId, cause));
+    }
+
+    /**
+     * This method is called when a NeonBee node has left the cluster.
+     *
+     * @param neonBee     the {@link NeonBee} instance
+     * @param hookContext the {@link HookContext}
+     * @param promise     {@link Promise} to completed the function.
+     */
+    @Hook(HookType.NODE_LEFT)
+    public void cleanup(NeonBee neonBee, HookContext hookContext, Promise<Void> promise) {
+        String clusterNodeId = hookContext.get(CLUSTER_NODE_ID);
+        LOGGER.info("Cleanup qualified names for node {}", clusterNodeId);
+        if (ClusterHelper.isLeader(neonBee.getVertx())) {
+            LOGGER.info("Cleaning registered qualified names ...");
+            unregister(neonBee, clusterNodeId).onComplete(promise)
+                    .onSuccess(unused -> LOGGER.info("Qualified names successfully cleaned up"))
+                    .onFailure(ignoredCause -> LOGGER.error("Failed to cleanup qualified names"));
+        }
+    }
+
+}
--- a/src/test/java/io/neonbee/internal/cluster/entity/ClusterEntityRegistryTest.java
+++ b/src/test/java/io/neonbee/internal/cluster/entity/ClusterEntityRegistryTest.java
@@ -1,0 +1,172 @@
+package io.neonbee.internal.cluster.entity;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.jupiter.api.DisplayName;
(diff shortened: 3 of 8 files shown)


s4heid (changes requested):
We need to come up with a solution that is reusable for other parts of NeonBee, like the `HealthCheckVerticle`. At the moment I don't see how `EntityVerticle.addToRegistrationList` and `EntityVerticle.unregister` can be reused. We should try to extract the reusable parts of the `EntityVerticle` into something like a `VerticleInspector`, that provides utilities to do the registering and unregistering of verticles. Maybe it's more effective to discuss these changes in a call.

src/main/java/io/neonbee/entity/EntityVerticle.java
```diff
@@ -254,30 +263,127 @@ private Future<Void> announceEntityVerticle(Vertx vertx) {
...
+
+    @VisibleForTesting
+    static Future<AsyncMap<Object, Object>> getNoteEntitiesMap(Vertx vertx) {
+        SharedDataAccessor sharedDataAccessor = new SharedDataAccessor(vertx, EntityVerticle.class);
+        return sharedDataAccessor.getAsyncMap("register");
+    }
+
+    private static List<Future> removeSharedMapEntries(Vertx vertx, JsonArray entries) {
+        AsyncMap<String, Object> sharedMap = NeonBee.get().getAsyncMap();
+        List<Future> futureList = new ArrayList<>(entries.size());
```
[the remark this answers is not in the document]
halber: This is not possible because the returned list is used to create a CompositeFuture with CompositeFuture.all. I mean, I could do this, but then I have to cast the list when I call CompositeFuture.all.

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
```diff
@@ -0,0 +1,70 @@
...
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.json.JsonArray;
+import io.vertx.junit5.Timeout;
+import io.vertx.junit5.VertxTestContext;
+
+@ExtendWith({ NeonBeeExtension.class })
+class UnregisterEntitiesTest {
+
+    @Test
+    @Timeout(value = 20, timeUnit = TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
halber: On my computer, it usually takes 5 seconds for the test to finish. But sometimes it takes 8 to 9 seconds. This is simply the upper limit. So lowering the value only increases the probability that the test will time out.

halber:
> We need to come up with a solution that is reusable for other parts of NeonBee, like the `HealthCheckVerticle`. At the moment I don't see how `EntityVerticle.addToRegistrationList` and `EntityVerticle.unregister` can be reused. We should try to extract the reusable parts of the `EntityVerticle` into something like a `VerticleInspector`, that provides utilities to do the registering and unregistering of verticles. Maybe it's more effective to discuss these changes in a call.

Currently I don't see the need to make this reusable. In fact, I don't think they should be used outside of the EntityVerticle class. These methods are used to change an internal (cluster-wide) state. Saying that, I should try to make addToRegistrationList private to avoid someone getting the idea to call this method.

src/main/java/io/neonbee/entity/EntityVerticle.java
```diff
@@ -250,34 +258,65 @@ private Future<Void> announceEntityVerticle(Vertx vertx) {
...
+                                        .onSuccess(event -> LOGGER.info(
+                                                "registered, cluster node ID: {}, Entity name: {}, qualified name: {} ",
+                                                clusterNodeId, name, qualifiedName));
+                            }
+                            return succeededFuture();
+                        })).onComplete(unused -> {
+                            LOGGER.debug("Release lock for cluster node ID: {}", clusterNodeId);
+                            lock.release();
+                        }));
+    }
```
pk-work, line 301: If you make *AsyncSharedMapRegistry* make also working for List<JsonObject> and not only for *String* you could reuse it. Maybe it is a little bit tricky, because in this case you directly add a whole list and not only one element into a list, but we can figure this out. I guess there is a pretty way to do this.

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
```diff
@@ -0,0 +1,88 @@
...
+import io.neonbee.internal.cluster.ClusterHelper;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonArray;
+
+/**
+ * Hooks for unregistering verticle models.
+ */
+public class UnregisterEntityModelHook {
```
pk-work, line 23:
```suggestion
public class UnregisterEntityVerticlesHook {
```

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
```diff
@@ -0,0 +1,88 @@
...
+     * @param nodeId hazelcast node ID
+     * @return Future
+     */
+    public static Future<Void> unregister(Vertx vertx, String nodeId) {
+        if (!vertx.isClustered()) {
+            return succeededFuture();
+        }
+
+        LOGGER.info("unregistering entity verticle models for node ID {} ...", nodeId);
+        Future<Void> unregisterFuture = EntityVerticle.getNodeEntitiesMap(vertx).compose(map -> map.get(nodeId))
```
pk-work, line 57: Where do you clean the **NodeEntitiesMap**? You only read from there in line 57, but when this information is used to clean up the "registry" with **removeSharedMapEntries()**, you could delete this information also in the  **NodeEntitiesMap**

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
```diff
@@ -0,0 +1,88 @@
...
+     * This method is called when a NeonBee node has left the cluster.
+     *
+     * @param neonBee     the {@link NeonBee} instance
+     * @param hookContext the {@link HookContext}
+     * @param promise     {@link Promise} to completed the function.
+     */
+    @Hook(HookType.NODE_LEFT)
+    public void cleanup(NeonBee neonBee, HookContext hookContext, Promise<Void> promise) {
+        String nodeId = hookContext.get(CLUSTER_NODE_ID);
+        LOGGER.info("cleanup models for node {}", nodeId);
```
pk-work, line 79: You don't clean up models, you clean up QualifiedNames for Entities.

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
```diff
@@ -0,0 +1,88 @@
...
+        LOGGER.info("Unregistering models on shutdown");
+        Vertx vertx = neonBee.getVertx();
+        String nodeId = ClusterHelper.getNodeId(neonBee.getVertx());
+        unregister(vertx, nodeId).onComplete(promise)
+                .onSuccess(unused -> LOGGER.info("models unregistered successfully"))
+                .onFailure(ignoredCause -> LOGGER.error("failed to unregister models on shutdown"));
+    }
+
+    /**
+     * Unregister the entity models for the node by ID.
```
pk-work, line 45: You unreigster QualifiedNames for Entities

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        })).compose(unused -> EntityVerticle.getNodeEntitiesMap(vertx)
+                .compose(map -> map.get(ClusterHelper.getNodeId(web.getVertx())))
+                .onSuccess(ja -> testContext.verify(() -> {
+                    JsonArray jsonArray = (JsonArray) ja;
+                    assertThat(jsonArray).hasSize(2);
+                }))).compose(unused -> UnregisterEntityModelHook.unregister(vertx, hazelcastNodeId))
+                .compose(unused -> web.getAsyncMap()
+                        .get(EntityVerticle.sharedEntityMapName(EntityVerticleUnregisterImpl.FQN_ERP_CUSTOMERS)))
+                .onSuccess(jsonArray -> testContext.verify(() -> {
+                    assertThat(jsonArray).isEqualTo(new JsonArray());
```
pk-work, line 64: isEmpty()?

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
```diff
@@ -0,0 +1,83 @@
...
+        Vertx vertx = web.getVertx();
+        String hazelcastNodeId = ClusterHelper.getNodeId(web.getVertx());
+
+        vertx.deployVerticle(new EntityVerticleUnregisterImpl()).onSuccess(unused -> testContext.verify(() -> {
+            assertThat(DeploymentHelper.isVerticleDeployed(vertx, EntityVerticleUnregisterImpl.class)).isTrue();
+        })).compose(unused -> EntityVerticle.getNodeEntitiesMap(vertx)
+                .compose(map -> map.get(ClusterHelper.getNodeId(web.getVertx())))
+                .onSuccess(ja -> testContext.verify(() -> {
+                    JsonArray jsonArray = (JsonArray) ja;
+                    assertThat(jsonArray).hasSize(2);
```
pk-work, line 59: you could check for the correct content not only the size

pk-work (dismissed):

pk-work:
Thanks that you working on this difficult topic

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    @Timeout(value = 20, timeUnit = TimeUnit.SECONDS)
+    @DisplayName("test unregistering entity models (Hazelcast cluster)")
+    void testHazelcastUnregisteringEntities(@NeonBeeInstanceConfiguration(activeProfiles = WEB, clustered = true,
+            clusterManager = HAZELCAST) NeonBee web, VertxTestContext testContext) {
+        testUnregisteringEntities(web, testContext);
+    }
+
+    private void testUnregisteringEntities(NeonBee web, VertxTestContext testContext) {
+        assertThat(isClustered(web)).isTrue();
```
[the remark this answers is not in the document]
halber: I don't see why it shouldn't work. testContext.verify is only required when I am in some async code to make the test fail. But at this point I am only failing the called test method. 
But I also tried it out and it does work.

src/main/java/io/neonbee/internal/cluster/ClusterHelper.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Get the cluster node ID.
+     *
+     * @param vertx {@link Vertx} instance
+     * @return the hazelcast cluster node ID
+     */
+    public static String getNodeId(Vertx vertx) {
+        return getClusterManager(vertx).map(ClusterManager::getNodeId).orElseThrow(() -> new IllegalStateException(
```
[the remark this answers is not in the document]
halber: Although I think this could be cached for Hazlecast or Infinispan, I can't make a statement about other ClusterManagers. But if it can and should be cached, for me the right place for it would be the implementation of the ClusterManager and not in NeonBee.

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
halber: good point!

src/main/java/io/neonbee/internal/cluster/ClusterHelper.java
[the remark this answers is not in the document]
pk-work: Okay, as long as, `getClusterManager(vertx).map(ClusterManager::getNodeId)` is quick, it is fine for me to not cache the NodeId directly.

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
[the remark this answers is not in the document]
pk-work: okay

src/test/java/io/neonbee/entity/UnregisterEntitiesTest.java
halber: I had to cast jsonArray to an iterable. But to be honest I think the generated error message is better this way.

src/main/java/io/neonbee/entity/UnregisterEntityModelHook.java
halber: Good catch! I didn't cleaned the map.

src/main/java/io/neonbee/internal/AsyncSharedMapRegistry.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Method that acquires a lock for the sharedMapKey and released the lock after the future is executed.
+     *
+     * @param sharedMapKey the shared map key
+     * @param future       {@link Future} to be secured by the lock
+     * @return the future
+     */
+    private Future<Void> lock(String sharedMapKey, Future<Void> future) {
+        return new SharedDataAccessor(vertx, AsyncSharedMapRegistry.class).getLock(sharedMapKey)
```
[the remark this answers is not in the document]
halber: Currently the default timeout is set. (10s)

src/main/java/io/neonbee/internal/cluster/ClusterHelper.java
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param vertx {@link Vertx} instance
+     * @return null or the {@link HazelcastClusterManager} instance
+     */
+    public static Optional<ClusterManager> getClusterManager(Vertx vertx) {
+        if (vertx instanceof VertxInternal) {
+            VertxInternal vertxInternal = (VertxInternal) vertx;
+            return Optional.ofNullable(vertxInternal.getClusterManager());
+        }
+        return Optional.empty();
```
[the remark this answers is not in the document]
halber: Correct would be:
`return Optional.ofNullable(vertx instanceof VertxInternal ? ((VertxInternal) vertx).getClusterManager() : null);`
but in my opinion, using the Ternaerer operator does hurt the readability of the code.

src/main/java/io/neonbee/entity/EntityVerticle.java
```diff
@@ -250,34 +258,65 @@ private Future<Void> announceEntityVerticle(Vertx vertx) {
         return entityTypeNames()
                 .map(entityTypeNames -> entityTypeNames != null ? entityTypeNames : Set.<FullQualifiedName>of())
                 .compose(entityTypeNames -> {
-                    AsyncMap<String, Object> sharedMap = NeonBee.get(vertx).getAsyncMap();
+                    AsyncSharedMapRegistry registry = new AsyncSharedMapRegistry(vertx);
```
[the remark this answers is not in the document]
halber: True, but it is only used twice. If I store it in a static variable, it stays in memory all the time.

s4heid (approved):
lgtm!

pk-work:
> lgtm!

Please first work on my comments.

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
```diff
@@ -0,0 +1,93 @@
...
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.Vertx;
+
+/**
+ * Hooks for unregistering verticle models.
+ */
+public class UnregisterEntityVerticlesHook {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(UnregisterEntityVerticlesHook.class);
```
s4heid, line 24: Just a personal preference, but would be nice to have all logs starting with the same case (either upper or lower). I think the majority is upper case.

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * This method is called when a NeonBee instance shutdown gracefully.
+     *
+     * @param neonBee     the {@link NeonBee} instance
+     * @param hookContext the {@link HookContext}
+     * @param promise     {@link Promise} to completed the function.
```
s4heid, line 31:
```suggestion
     * @param promise     {@link Promise} to complete the function.
```

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    @Hook(HookType.BEFORE_SHUTDOWN)
+    public void unregisterOnShutdown(NeonBee neonBee, HookContext hookContext, Promise<Void> promise) {
+        LOGGER.info("Unregistering models on shutdown");
+        Vertx vertx = neonBee.getVertx();
+        String nodeId = ClusterHelper.getNodeId(neonBee.getVertx());
```
s4heid, line 37:
```suggestion
        String nodeId = ClusterHelper.getNodeId(vertx);
```

you already have vertx in a local variable (line above)

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * Unregister the entity qualified names for the node by ID.
+     *
+     * @param vertx  The Vert.x instance
+     * @param nodeId hazelcast node ID
+     * @return Future
+     */
+    public static Future<Void> unregister(Vertx vertx, String nodeId) {
```
s4heid, line 50: What do think about passing NeonBee instead of vertx + nodeId to this public method? We would need one parameter less.

src/test/java/io/neonbee/internal/cluster/entity/ClusterEntityRegistryTest.java
```diff
@@ -0,0 +1,171 @@
...
+        ClusterEntityRegistry registry = new TestClusterEntityRegistry(vertx, REGISTRY_NAME);
+        registry.register(KEY, VALUE).compose(unused -> registry.get(KEY)).onSuccess(jsonArray -> context.verify(() -> {
+            assertThat(jsonArray).isNotNull();
+            assertThat(jsonArray.contains(VALUE)).isTrue();
+            context.completeNow();
+        })).onFailure(context::failNow);
+    }
+
+    @Test
+    @Timeout(value = 1, unit = TimeUnit.SECONDS)
```
s4heid, line 63: display name?

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
```diff
@@ -0,0 +1,93 @@
...
+
+        LOGGER.info("unregistering entity verticle models for node ID {} ...", nodeId);
+        Future<Void> unregisterFuture = clusterEntityRegistry.unregisterNode(nodeId);
+
+        unregisterFuture
+                .onSuccess(unused -> LOGGER.info("unregistered entity verticle models for node ID {} ...", nodeId))
+                .onFailure(cause -> LOGGER.error("failed to unregistered entity verticle models for node ID {} ...",
+                        nodeId, cause));
+
+        return unregisterFuture;
```
s4heid, line 71: why this extra return, and not returning line 66 already?

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
halber: Good point, never paid attention to it.

src/main/java/io/neonbee/internal/cluster/entity/UnregisterEntityVerticlesHook.java
halber: Passing NeonBee instead of vertx makes sense. However, I still need the nodeId as a parameter. See UnregisterEntityVerticlesHook#cleanup

halber: I understand the confusion now. It is not the NeonBee#getNodeId here. The node ID of the cluster node is needed. I renamed the corresponding parameters all to clusterNodeId to make it clearer by the name what is needed.

halber: Otherwise, I need to introduce a class to make it typesafe. But I did not consider it necessary.

s4heid: No that's fine! thanks for making this clear

s4heid (approved):

PR #165, opened 2022-08-10
feat: extend DataException and return DataException via eventbus

dapeng-wang:
Currently DataException only allows propagation of one error message. The intention is to add an failureDetail message to be able to propagate source error messages to the invoker, but should not be exposed to the consumer. A message codec for DataException is provided to allow serialization/deserialization of DataException objects via the event bus.

### DIFF
--- a/src/main/java/io/neonbee/data/DataException.java
+++ b/src/main/java/io/neonbee/data/DataException.java
@@ -1,9 +1,15 @@
 package io.neonbee.data;
 
 import static io.neonbee.internal.helper.StringHelper.EMPTY;
+import static java.util.Objects.requireNonNull;
 
+import java.util.Collections;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
 
 @SuppressWarnings({ "OverrideThrowableToString", "checkstyle:JavadocVariable" })
 public class DataException extends RuntimeException {
@@ -20,6 +26,8 @@ public class DataException extends RuntimeException {
     private static final long serialVersionUID = 1L;
 
     private final int failureCode;
+
+    private final Map<String, Object> failureDetail;
 
     /**
      * Create a DataException.
@@ -53,8 +61,22 @@ public class DataException extends RuntimeException {
      * @param message     the failure message
      */
     public DataException(int failureCode, String message) {
+        this(failureCode, message, Map.of());
+    }
+
+    /**
+     * Create a DataException.
+     *
+     * @param failureCode   the failure code
+     * @param message       the failure message
+     * @param failureDetail the failure detail message to be propagated to the invoker, but should not be exposed to the
+     *                      consumer. Must be compatible to {@link JsonObject#JsonObject(Map)}. The type of the passed
+     *                      object must be either: {@link String} or {@link JsonObject} or {@link JsonArray}
+     */
+    public DataException(int failureCode, String message, Map<String, Object> failureDetail) {
         super(message);
         this.failureCode = failureCode;
+        this.failureDetail = requireNonNull(failureDetail);
     }
 
     /**
@@ -66,9 +88,18 @@ public class DataException extends RuntimeException {
         return failureCode;
     }
 
+    /**
+     * Get the failure detail message.
+     *
+     * @return the failure detail
+     */
+    public Map<String, Object> failureDetail() {
+        return Collections.unmodifiableMap(failureDetail);
+    }
+
     @Override
     public int hashCode() {
-        return Objects.hash(failureCode, getMessage());
+        return Objects.hash(failureCode, getMessage(), failureDetail);
     }
 
     @Override
@@ -83,11 +114,14 @@ public class DataException extends RuntimeException {
             return false;
         }
         DataException other = (DataException) obj;
-        return failureCode == other.failureCode && getMessage().equals(other.getMessage());
+        return failureCode == other.failureCode
+                && Optional.ofNullable(getMessage()).orElse(EMPTY)
+                        .equals(Optional.ofNullable(other.getMessage()).orElse(EMPTY))
+                && failureDetail().equals(other.failureDetail());
     }
 
     @Override
     public String toString() {
-        return "(" + failureCode + ") " + Optional.ofNullable(getMessage()).orElse(EMPTY);
+        return "(" + failureCode + ")" + Optional.ofNullable(getMessage()).map(msg -> " " + msg).orElse(EMPTY);
     }
 }
--- a/src/main/java/io/neonbee/data/DataVerticle.java
+++ b/src/main/java/io/neonbee/data/DataVerticle.java
@@ -115,14 +115,22 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
                         LOGGER.correlateWith(context).debug("Received event bus reply");
 
                         if (asyncReply.succeeded()) {
-                            DataContext responseDataContext =
-                                    decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER));
-                            context.setData(
-                                    Optional.ofNullable(responseDataContext).map(DataContext::data).orElse(null));
-                            context.mergeResponseData(Optional.ofNullable(responseDataContext)
-                                    .map(DataContext::responseData).orElse(null));
-                            return succeededFuture(asyncReply.result().body());
-
+                            U body = asyncReply.result().body();
+                            if (body instanceof DataException) {
+                                if (LOGGER.isWarnEnabled()) {
+                                    LOGGER.correlateWith(context).warn("Received a event bus reply failure from {}",
+                                            qualifiedName, (DataException) body);
+                                }
+                                return failedFuture((DataException) body);
+                            } else {
+                                DataContext responseDataContext =
+                                        decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER));
+                                context.setData(
+                                        Optional.ofNullable(responseDataContext).map(DataContext::data).orElse(null));
+                                context.mergeResponseData(Optional.ofNullable(responseDataContext)
+                                        .map(DataContext::responseData).orElse(null));
+                                return succeededFuture(asyncReply.result().body());
+                            }
                         } else {
                             Throwable cause = asyncReply.cause();
                             if (LOGGER.isWarnEnabled()) {
@@ -412,7 +420,7 @@ public abstract class DataVerticle<T> extends AbstractVerticle implements DataAd
                             }
 
                             if (cause instanceof DataException) {
-                                message.fail(((DataException) cause).failureCode(), cause.getMessage());
+                                message.reply(cause);
                             } else {
                                 message.fail(FAILURE_CODE_PROCESSING_FAILED,
                                         "Processing of message failed. " + cause.getMessage());
--- a/src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
+++ b/src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
@@ -1,0 +1,70 @@
+package io.neonbee.internal.codec;
+
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+import io.neonbee.data.DataException;
+import io.netty.util.CharsetUtil;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.MessageCodec;
+import io.vertx.core.json.JsonObject;
+
+public class DataExceptionMessageCodec implements MessageCodec<DataException, DataException> {
+
+    @Override
+    public void encodeToWire(Buffer buffer, DataException exception) {
+        buffer.appendInt(exception.failureCode());
+
+        String msgEncoded = exception.getMessage();
+        if (msgEncoded != null) {
+            byte[] msgBytes = msgEncoded.getBytes(CharsetUtil.UTF_8);
+            buffer.appendInt(msgBytes.length);
+            buffer.appendBytes(msgBytes);
+        } else {
+            buffer.appendInt(-1);
+        }
+
+        Buffer failureDetailEncoded = new JsonObject(exception.failureDetail()).toBuffer();
+        buffer.appendInt(failureDetailEncoded.length());
+        buffer.appendBuffer(failureDetailEncoded);
+    }
+
+    @Override
+    public DataException decodeFromWire(int pos, Buffer buffer) {
+        int tmpPos = pos;
+        int failureCode = buffer.getInt(tmpPos);
+        tmpPos += Integer.BYTES;
+
+        int msgLength = buffer.getInt(tmpPos);
+        tmpPos += Integer.BYTES;
+        String message =
+                msgLength != -1 ? new String(buffer.getBytes(tmpPos, tmpPos + msgLength), CharsetUtil.UTF_8) : null;
+        tmpPos += msgLength != -1 ? msgLength : 0;
+
+        int failureDetailLength = buffer.getInt(tmpPos);
+        tmpPos += Integer.BYTES;
+
+        Buffer buff = buffer.getBuffer(tmpPos, tmpPos + failureDetailLength);
+        Map<String, Object> failureDetail =
+                buff.toJsonObject().stream().collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+
+        return new DataException(failureCode, message, failureDetail);
+    }
+
+    @Override
+    public DataException transform(DataException exception) {
+        return exception;
+    }
+
+    @Override
+    public String name() {
+        return "dataexception";
+    }
+
+    @Override
+    public byte systemCodecID() {
+        return -1;
+    }
+
+}
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -42,6 +42,7 @@ import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import io.neonbee.config.HealthConfig;
 import io.neonbee.config.NeonBeeConfig;
 import io.neonbee.config.ServerConfig;
+import io.neonbee.data.DataException;
 import io.neonbee.data.DataQuery;
 import io.neonbee.entity.EntityModelManager;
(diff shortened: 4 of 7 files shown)


pk-work:
Whats the benefit of "internalMessage"? Where or how is it used? Some times the intention of a small change is clear, especially if the change contains a test.

At least for me the intention it is not clear, so maybe you can write a short Explanatory Text in the commit message to explain it.

dapeng-wang:
> Whats the benefit of "internalMessage"? Where or how is it used? Some times the intention of a small change is clear, especially if the change contains a test.
> 
> At least for me the intention it is not clear, so maybe you can write a short Explanatory Text in the commit message to explain it.

Currently DataException is only able to transport one error message. In case of W7 it returns only 400 Bad Request. The internal error message from W7 is completely swallowed. 

After discussion with Kristian we decided to add an internalMessage to transport detailed error message to the invoker, which will not be exposed to the external consumer.

kristian:
This is a good change and necessary, however what I am little unsure is about the naming... "internalMessage", hmm.

I know I have seen a better naming in some other open source projects for this type of "technical" error message, but I do not recall... Let me think about this for a second.

pk-work:
Okay, I agree that this is necessary when you have a generic Verticle, where you cannot respond with a pre-defined error code.

But I'd really appreciate if you can add a test for this.

pk-work:
Maybe `detailedMessage` is an alternative to `internalMessage`. If you send back an Exception via the Eventbus, wouldn't you need an Message Codec for that?`

What came just in my mind is, whats the difference between message and detailedMessage? If the "normal" message just a generi thing like "Bad Request", you could also use an error code for this and use the message field for the detailed message.

ada89:
> Maybe `detailedMessage` is an alternative to `internalMessage`. If you send back an Exception via the Eventbus, wouldn't you need an Message Codec for that?`
> 
> What came just in my mind is, whats the difference between message and detailedMessage? If the "normal" message just a generic thing like "Bad Request", you could also use an error code for this and use the message field for the detailed message.

Why we did choose to have a new field for the internal/detail message was the fact that we don't want this message to be exposed to the consumer.

pk-work:
If you choose an error code below 100 or above 1000, it won't be published to the consumer. Maybe this already solves your problem?

https://github.com/SAP/neonbee/blob/3b48202cb0e4ff7d3433562589f28c53c322acb0/src/main/java/io/neonbee/internal/handler/DefaultErrorHandler.java#L77

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -130,7 +130,7 @@ public MessageCodec<T, T> getMessageCodec() {
     @Override
     public void init(Vertx vertx, Context context) {
         super.init(vertx, context);
-        JsonObject metrics = getMetricsConfig(NeonBee.get().getConfig().getMetricsConfig());
+        JsonObject metrics = getMetricsConfig(NeonBee.get(vertx).getConfig().getMetricsConfig());
```
s4heid, line 133: This line is actually quite important although it has nothing to do with the original change described in the commit message. Maybe it's worth to separate this into a different commit in order to make it clear.

s4heid: Mentioning this as this line bite me today, too when trying to upgrade to the latest vert.x version.

ada89: I agree. I dropped a line to Benni about it once I saw he updated core and the some of the tests were failing. 

I noticed some regressions due to this while working on some feature in core last week. I discussed with Michael and we agreed to pack it in this commit, thinking this one will be merged soon (I agree it was not the best decision and we should do it in a separate commit).

kristian:
Ok, as promised I again made my mind up a little bit about the naming. Maybe to make it a little more concrete:

What we mean with this message is some kind of message which is useful for logging / technical purpose, but is not to be exposed to end users maybe for security purpose, so also our endpoints should not expose this message.

What I don't like so much about the term "internal" (as in internalMessage) is that it is unclear to me, so the purpose of this message if it is labeled "intern" is not so clear to me, it rather sounds like it is a "internal" thing to the DataException, but that is not the case, it is a error description which might contain "too much" detail, to be exposed.

More or less what I am leaning toward is calling the message rather "confidential" or "confied" or "protected" or "supressed", so some word that makes it more clear that it is not only "internal" but that it might contain information that is not to be exposed in order to not cause a security leak.

However I also now read about to topic a little bit... I am getting more and more unsure about whether to at all go into this direction: General opinion here from the Java Community: Java Exception messages should _never_ be exposed to any end-users at all... meaning that "message" would stay the right place to put information about failures. As for this example, if further information of "what" failed should be exposed the "cause" of the exception should point in that direction. In our case, where the event bus communication is "in between" this might get a little bit more complex I agreee... the question is how we deal with those "boundaries"...

The question is: Is the problem that DataException "swallows" the internal exception and we should improve that, or is it that we should include private information in a DataExceptions message and then rather deal with it upstream to never expose the message how it is normally done in Java?

ada89:
> 



> Ok, as promised I again made my mind up a little bit about the naming. Maybe to make it a little more concrete:
> 
> What we mean with this message is some kind of message which is useful for logging / technical purpose, but is not to be exposed to end users maybe for security purpose, so also our endpoints should not expose this message.
> 
> What I don't like so much about the term "internal" (as in internalMessage) is that it is unclear to me, so the purpose of this message if it is labeled "intern" is not so clear to me, it rather sounds like it is a "internal" thing to the DataException, but that is not the case, it is a error description which might contain "too much" detail, to be exposed.
> 
> More or less what I am leaning toward is calling the message rather "confidential" or "confied" or "protected" or "supressed", so some word that makes it more clear that it is not only "internal" but that it might contain information that is not to be exposed in order to not cause a security leak.
> 
> However I also now read about to topic a little bit... I am getting more and more unsure about whether to at all go into this direction: General opinion here from the Java Community: Java Exception messages should _never_ be exposed to any end-users at all... meaning that "message" would stay the right place to put information about failures. As for this example, if further information of "what" failed should be exposed the "cause" of the exception should point in that direction. In our case, where the event bus communication is "in between" this might get a little bit more complex I agreee... the question is how we deal with those "boundaries"...
> 
> The question is: Is the problem that DataException "swallows" the internal exception and we should improve that, or is it that we should include private information in a DataExceptions message and then rather deal with it upstream to never expose the message how it is normally done in Java?

Thank you for the input @kristian . 

The problem according to my understanding is that DataException "swallows" this internal exception and we would like to improve that.

ada89 (dismissed):
approved on behalf of DaPeng (as he is the author of the PR, he is not able to approve it)

src/main/java/io/neonbee/data/DataException.java
```diff
@@ -53,8 +63,21 @@ public DataException(int failureCode) {
...
+
+    /**
+     * Create a DataException.
+     *
+     * @param failureCode   the failure code
+     * @param message       the failure message
+     * @param failureDetail the failure detail message to be propagated to the invoker, but should not be exposed to the
+     *                      consumer
+     */
+    public DataException(int failureCode, String message, Map<String, Buffer> failureDetail) {
```
kristian, line 77: Can we use a map of <String, Object> here and make it clear in JavaDoc that the Object must be compatible to the Vert.x event bus codecs which are registered? That way somebody could also just provide a <String, String> as details and doesn't necesarry need to convert it to buffer first and then loose all the context information...

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -343,11 +343,20 @@ public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataCo
...
                         if (asyncReply.succeeded()) {
-                            context.setData(Optional
-                                    .ofNullable(
-                                            decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER)))
-                                    .map(DataContext::data).orElse(null));
-                            return succeededFuture(asyncReply.result().body());
+                            U body = asyncReply.result().body();
+                            if (body instanceof DataException) {
+                                if (LOGGER.isWarnEnabled()) {
+                                    LOGGER.correlateWith(context).warn("Failed to receive event bus reply from {}",
```
kristian, line 349: I think we can get a little more specific here, because it no longer means that the "receive failed", but maybe something along those likes of "Received a event bus reply failure from {}"

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -343,11 +343,20 @@ public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataCo
...
+                                            qualifiedName, (DataException) body);
+                                }
+                                return failedFuture(mapException((DataException) body));
```
kristian, line 352: We can spare the mapException here, because we already validated that this is a DataException here

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,38 @@
...
+import io.neonbee.data.DataException;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.MessageCodec;
+import io.vertx.core.json.JsonObject;
+
+public class DataExceptionMessageCodec implements MessageCodec<DataException, DataException> {
+
+    @Override
+    public void encodeToWire(Buffer buffer, DataException body) {
+        JsonObject json = JsonObject.mapFrom(body);
```
kristian, line 12: Uhhh, really? Do we rely encoding this class with Jackson here?

Are we sure that this is intended? I mean both Vert.x documentation and also Jackson documentation clearly mentions that the mapFrom / encode they do is intended for POJO's only:

https://vertx.io/docs/apidocs/io/vertx/core/json/Json.html#encode-java.lang.Object-

Our DataException is clearly NO POJO! I think we have to invest in an own encoding here, this is prone to failure!

kristian (changes requested):
See comments attached

kristian:
Maybe just one more idea, a map of <String, Object> can be also used to initialize a JsonObject(Map), so we can send it via the event bus more easily

kristian:
Wow, that was a quick impl. of that codec!!

src/main/java/io/neonbee/data/DataVerticle.java
```diff
@@ -343,11 +343,20 @@ public static <U> Future<U> requestData(Vertx vertx, DataRequest request, DataCo
...
                         if (asyncReply.succeeded()) {
-                            context.setData(Optional
-                                    .ofNullable(
-                                            decodeContextFromString(asyncReply.result().headers().get(CONTEXT_HEADER)))
-                                    .map(DataContext::data).orElse(null));
-                            return succeededFuture(asyncReply.result().body());
+                            U body = asyncReply.result().body();
+                            if (body instanceof DataException) {
+                                if (LOGGER.isWarnEnabled()) {
+                                    LOGGER.correlateWith(context).warn("Received a event bus reply failure from {}}",
```
kristian, line 349: One } too many at the end :)

src/main/java/io/neonbee/data/DataVerticle.java
ada89: done

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,117 @@
...
+        buffer.appendInt(body.failureCode());
+        if (!Strings.isNullOrEmpty(body.getMessage())) {
+            appendObjectToBuffer(body.getMessage(), buffer);
+        }
+        if (!body.failureDetail().isEmpty()) {
+            failureBody.keySet().forEach(key -> {
+                appendObjectToBuffer(key, buffer);
+                appendObjectToBuffer(failureBody.get(key), buffer);
+            });
+        }
```
kristian, line 27: How about something along those lines: First I think body should be renamed `exception` or `ex` to make it more clear what is being encoded. Then:

```java
buffer.appendInt(ex.failureCode())

String msg = ex.getMessage();
if (msg != null) {
  byte[] strBytes = s.getBytes(CharsetUtil.UTF_8);
  buffer.appendInt(strBytes.length);
  buffer.appendBytes(strBytes);
} else {
  buffer.appendInt(-1);
}

Buffer encDetail = new JsonObject(ex.failureDetail()).toBuffer(); // we can also put this in the JavaDoc of the DataException, that the failureMap must be compatible to the JsonObject constructor (you can link it with {@link JsonObject(Map)} so it is clear to the end-user
buffer.appendInt(encDetail.length());
buffer.appendBuffer(encDetail);
```

Then the decoding can be also much simplified! 👍

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,117 @@
...
+        } else if (obj instanceof JsonArray) {
+            buffer.appendInt(2); // 2 means JsonArray
+            encoded = ((JsonArray) obj).toBuffer();
+        } else {
+            throw new UnsupportedOperationException(
+                    "Serialization of type " + Object.class.getName() + " is not supported");
+        }
+        buffer.appendInt(encoded.length());
+        buffer.appendBuffer(encoded);
+    }
```
kristian, line 47: remove

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,117 @@
...
+                case 2:
+                    failureDetailValue = new JsonArray(failureDetailBodyAsBuffer);
+                    break;
+                default:
+                    throw new UnsupportedOperationException(
+                            "Deserialization of type " + Object.class.getName() + " is not supported");
+                }
+            }
+            failureDetail = Map.of(failureDetailKey, failureDetailValue);
+        }
```
kristian, line 98:
```java
int tmpPos = pos; // such a stupid rule that we have to put this into an own variable here, if we have a perfectly good inbound variable already ;)
int failureCode = buffer.getInt(tmpPos);
tmpPos += 4; // this is how Vert.x does it... think we should do it the same, see https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/eventbus/impl/codecs/ReplyExceptionMessageCodec.java#L44

int strLength = buffer.getInt(tmpPos);
tmpPos += 4;
String message = strLength == -1 ? null : new String(buffer.getBytes(tmpPos, tmpPos + strLength), CharsetUtil.UTF_8);
tmpPos += strLength;

int detailLength = buffer.getInt(tmpPos);
Map<String, Object> failureDetail = buffer.getBuffer(tmpPos, tmpPos + detailLength).toJsonObject().getMap();

return new DataException(failureCode, message, failureDetail);
```

kristian (changes requested):

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
ada89: Hi Kristian,

thanks for the concrete proposal! I have implemented it and there seems to be one issue:
(according to my understanding) when decoding the failureDetail from the buffer, in case the value of the map is i.e. JsonObject, it get decoded to a LinkedHashMap instead of a JsonObject. The same for a JsonArray -> decoded to an ArrayList. Basically I noticed that the unit tests were failing when in the map I had a JsonObject or a JsonArray. 

The equals method of the DataException checks that the equality between 2 objects, with the default equals method of the Map -  I would say this is right, or?

I have attached a screenshot of the actual and expected objects from the failing test to clarify what I summed up above. 

Screenshot 2022-08-25 at 21 44 40

kristian:
lgtm as soon as the build is green I think we can merge this one :)

kristian:
Ah the only thing that maybe in the DataException constructor we shall mention the Json Map constructor, so end users know what to pass in and what not

kristian:
great! Thanks so much @ada89 ! lgtm as soon as green :) have a great evening!

s4heid (dismissed):

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,79 @@
...
+            byte[] msgBytes = msgEncoded.getBytes(CharsetUtil.UTF_8);
+            buffer.appendInt(msgBytes.length);
+            buffer.appendBytes(msgBytes);
+        } else {
+            buffer.appendInt(-1);
+        }
+
+        if (exception.failureDetail().isEmpty()) {
+            buffer.appendInt(-1);
+        } else {
```
kristian, line 31: Is this required? An empty JsonObject should be encoded as {}, so we do not need special logic for it :)

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
```diff
@@ -0,0 +1,79 @@
...
+
+        int msgLength = buffer.getInt(tmpPos);
+        tmpPos += Integer.BYTES;
+        String message =
+                msgLength != -1 ? new String(buffer.getBytes(tmpPos, tmpPos + msgLength), CharsetUtil.UTF_8) : null;
+        tmpPos += msgLength;
+
+        int failureDetailLength = buffer.getInt(tmpPos);
+        tmpPos += Integer.BYTES;
+        Map<String, Object> failureDetail = failureDetailLength != -1
```
kristian, line 53: Same as above

src/main/java/io/neonbee/internal/codec/DataExceptionMessageCodec.java
kristian: Mhm yes I see the problem... There is no "real" solution for this because internally JsonObject is handled as Map and JsonArray is handled as List... and JsonObject/Array has handling in the getter method for that. So what it was "originally" can not be determined. As in my humble oppinion we can assume that most people will be dealing with JsonObject / JsonArray rather than with plain lists etc. for the failureDetail, my suggestion would be to use the stream method instead, when decoding the result, so instead of doing:

```java
buffer.getBuffer(tmpPos, tmpPos + detailLength).toJsonObject().getMap();
```

Do a:

```java
buffer.getBuffer(tmpPos, tmpPos + detailLength).toJsonObject().stream().collect(Collectors.toMap(Entry::getKey, Entry::getValue));
```

That should trigger the internal conversion by Vert.x into JsonObject / JsonArray. Hope this helps! :)

PR #166, opened 2022-08-12
feat: SFMBBB-305 entity model manager controls shared model definition

mkhazanovych:
Every node pushes its entity model definitions in shared memory and is able to reload them from shared memory.

### DIFF
--- a/src/main/java/io/neonbee/entity/EntityModelManager.java
+++ b/src/main/java/io/neonbee/entity/EntityModelManager.java
@@ -3,13 +3,9 @@ package io.neonbee.entity;
 import static io.vertx.core.Future.failedFuture;
 import static io.vertx.core.Future.succeededFuture;
 
-import java.util.Collections;
-import java.util.ConcurrentModificationException;
-import java.util.Map;
-import java.util.NoSuchElementException;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.olingo.server.api.OData;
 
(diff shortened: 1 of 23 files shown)


src/main/java/io/neonbee/entity/EntityModelManager.java
```diff
@@ -357,102 +230,155 @@ public static Future<Map<String, EntityModel>> reloadModels(NeonBee neonBee) {
...
-    public Future<Map<String, EntityModel>> registerModels(EntityModelDefinition modelDefinition) {
-        return externalModelDefinitions.add(modelDefinition) ? reloadModels() : getSharedModels();
+    public Future<Void> pushChanges(Vertx vertx) {
+        LOGGER.info("pushing entity models changes in shared memory");
+        final List<Future> futures = new ArrayList<>(externalModelDefinitions.size());
+        return getLock(vertx).compose(lock -> getAsyncMap(vertx).compose(m -> {
+            externalModelDefinitions.forEach(e -> futures.add(m.putIfAbsent(e.toBuffer(), getNodeId(vertx))));
+            LOGGER.info("entity models written into shared memory by node id " + getNodeId(vertx) + " = "
+                    + externalModelDefinitions);
+            return CompositeFuture.all(futures).compose(result -> {
```
halber, line 275: Shouldn't we use join instead of all to wait for all futures even if a future fails?

src/main/java/io/neonbee/entity/EntityModelManager.java
```diff
@@ -357,102 +230,155 @@ public static Future<Map<String, EntityModel>> reloadModels(NeonBee neonBee) {
...
+                if (!externalModelDefinitions.contains(modelDefinition)) {
+                    diff.add(modelDefinition);
+                    externalModelDefinitions.add(modelDefinition);
+                } else {
+                    LOGGER.info("obtained remote model definition " + modelDefinition + " is already known");
+                }
+            });
+            LOGGER.info("obtained remote model definitions " + diff);
+            return succeededFuture();
+        }).compose(dummy -> succeededFuture(diff));
```
halber, line 332:
```suggestion
return getAsyncMap(NeonBee.get().getVertx()).compose(AsyncMap::keys).compose(m -> {
            LOGGER.info("obtaining remote model definitions iterating over async map ");
            Set<EntityModelDefinition> diff = new HashSet<>(m.size());
            m.forEach(buffer -> {
                EntityModelDefinition modelDefinition = EntityModelDefinition.fromBuffer(buffer);
                if (externalModelDefinitions.contains(modelDefinition)) {
                    LOGGER.info("obtained remote model definition " + modelDefinition + " is already known");
                } else {
                    diff.add(modelDefinition);
                    externalModelDefinitions.add(modelDefinition);
                }
            });
            LOGGER.info("obtained remote model definitions " + diff);
            return succeededFuture(diff);
        });
```

halber (changes requested):

Issue #148, opened 2022-07-12
Make cluster manager choosable


pk-work:
At the moment NeonBee is using Hazelcast as a cluster manager, because it is small, fast and can be embedded.

But when you want to use the following features, you need to pay [1]:
 - Authentication: To ensure that not everyone can add a Node to your cluster
 - Encryption: To ensure that not everyone can read the inter cluster node communication.

If a project requires these features but don't want to buy a Hazelcast license, the project can't use NeonBee. In my opinion this is a problem.

There is an Infinispan Cluster Manager [2] that provides the same features as Hazelcast, but in addition it also offers Authentication and Encryption. Infinispan is licensed under the Apache License 2.0.

I'd like to replace Hazelcast by Infinispan or at least offer the possibility to choose another Cluster Manager.

**Impact:** When we introduced the `HazelcastClusterHealthCheck` [4], we added a hard dependency to Hazelcast in our productive code. We need to check how we can make this more abstract, or add also a `InfinispanClusterHealthCheck`.

[1] https://hazelcast.com/product-features/imdg-comparison/
[2] https://vertx.io/docs/vertx-infinispan/java/
[3] https://infinispan.org/features/
[4] https://github.com/SAP/neonbee/blob/main/src/main/java/io/neonbee/health/HazelcastClusterHealthCheck.java

pk-work:
I'd like to start to work on this issue. Before I come up with a Concept / PoC we need to clarify one important question:

- Do we want to have separate libraries for Hazelcast and Infinispan, we need to put on the classpath?

or

- Do we want to have Hazelcast and Infinispan dependencies in NeonBee and switch the ClusterManager via an option?

s4heid:
First of all, I think this change is a good one. These are essential features, many consumers probably would like to use.

I think replacing the current cluster manager is not a feasible option (at least not in short-term). Consumers of NeonBee might already have Hazelcast specific configuration as part of their installations and relying on Hazelcast features (which are potentially not available with Infinispan).

* > Do we want to have Hazelcast and Infinispan dependencies in NeonBee and switch the ClusterManager via an option?
  
  In my opinion, **yes**. Instead of replacing we can keep hazelcast as a default (without violating the "we choose, you use" motto of NeonBee), but add a `--cluster-manager` option to the `NeonBeeOptions` which allows to choose between `hazelcast` (default) and `infinispan`.

[...]

PR #182, opened 2022-09-27
WIP: feat: add Infinispan ClusterManager

pk-work:


pk-work:
closes #148

src/main/java/io/neonbee/NeonBee.java
```diff
@@ -260,12 +256,17 @@ static Future<NeonBee> create(Function<VertxOptions, Future<Vertx>> vertxFactory
...
-            vertxOptions.setClusterManager(clusterManagerFactory.apply(options));
-        }
+
+        Future<VertxOptions> loadClusterManager = Future.future(p -> {
+            if (options.isClustered()) {
+                p.handle(clusterManagerFactory.apply(options).map(vertxOptions::setClusterManager));
+            } else {
+                p.complete(vertxOptions);
+            }
+        });
```
halber, line 266: Just a thought: we could remove this if block by providing a NO_CLUSTER_FACTORY ClusterManagerFactory that simply returns null. But then we have to make this decision in the NeonBeeOptions.

halber (commented):
Looks very good. I suggested just one small improvement in ClusterManagerFactory.

src/main/java/io/neonbee/NeonBee.java
pk-work: We will not do this as discussed

halber (approved):
LGTM

PR #185, opened 2022-10-06
fix(health): make health check collection more resilient

s4heid:
Whenever a health check verticle did not answer, a failed future was returned by the health check registry in the attempt of collecting all health check data. Especially in the clustered case this might happen quite often due to restarts caused by a deployment, for instance. In that case an error would be reported, although all relevent data might be available.

Therefore, this change makes the data collection more resilient by respecting that some verticles might not be able to answer.

Once the de-registering of health check verticles from the shared map is available/implemented, it should be less likely that this situation happens, but still possible.

### DIFF
--- a/src/main/java/io/neonbee/health/HealthCheckRegistry.java
+++ b/src/main/java/io/neonbee/health/HealthCheckRegistry.java
@@ -173,17 +173,27 @@ public class HealthCheckRegistry {
         List<Future<JsonObject>> asyncCheckResults =
                 getHealthChecks().values().stream().map(hc -> hc.result().map(CheckResult::toJson)).collect(toList());
 
-        return AsyncHelper.allComposite(asyncCheckResults).map(resolvedCheckResults -> resolvedCheckResults.list()
-                .stream().map(JsonObject.class::cast).peek(result -> result.remove("outcome")).collect(toList()));
+        return AsyncHelper.joinComposite(asyncCheckResults).transform(v -> {
+            List<JsonObject> allSucceeded = asyncCheckResults.stream().filter(Future::succeeded).map(Future::result)
+                    .peek(result -> result.remove("outcome")).collect(toList());
+
+            return Future.succeededFuture(allSucceeded);
+        });
     }
 
     private Future<List<JsonObject>> getClusteredHealthCheckResults(DataContext dataContext) {
         return NeonBee.get(vertx).getAsyncMap().get(SHARED_MAP_KEY)
                 .map(qualifiedNames -> (qualifiedNames != null ? (JsonArray) qualifiedNames : new JsonArray()))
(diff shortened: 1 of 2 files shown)


src/main/java/io/neonbee/health/HealthCheckRegistry.java
```diff
@@ -173,17 +173,27 @@ Future<List<JsonObject>> getLocalHealthCheckResults() {
         List<Future<JsonObject>> asyncCheckResults =
                 getHealthChecks().values().stream().map(hc -> hc.result().map(CheckResult::toJson)).collect(toList());
 
-        return AsyncHelper.allComposite(asyncCheckResults).map(resolvedCheckResults -> resolvedCheckResults.list()
-                .stream().map(JsonObject.class::cast).peek(result -> result.remove("outcome")).collect(toList()));
+        return AsyncHelper.joinComposite(asyncCheckResults).transform(v -> {
+            List<JsonObject> allSucceeded = asyncCheckResults.stream().filter(Future::succeeded).map(Future::result)
+                    .map(JsonObject.class::cast).peek(result -> result.remove("outcome")).collect(toList());
```
halber, line 178: I don't see why we need to convert the result to a JsonObject.
```suggestion
                    .peek(result -> result.remove("outcome")).collect(toList());
``

File: src/main/java/io/neonbee/health/HealthCheckRegistry.java [hunk 2]
```diff
@@ -173,17 +173,27 @@ Future<List<JsonObject>> getLocalHealthCheckResults() {
...
-                .compose(qualifiedNames -> AsyncHelper.allComposite(sendDataRequests(qualifiedNames, dataContext)))
-                .map(resolvedRequests -> {
-                    return resolvedRequests.list().stream().map(JsonArray.class::cast).flatMap(JsonArray::stream)
-                            .map(JsonObject.class::cast).collect(toList());
+                .compose(qualifiedNames -> {
+                    List<Future<JsonArray>> asyncCheckResults = sendDataRequests(qualifiedNames, dataContext);
+
+                    return AsyncHelper.joinComposite(asyncCheckResults).transform(v -> {
+                        List<JsonObject> allSucceeded = asyncCheckResults.stream().filter(Future::succeeded)
+                                .map(Future::result).map(JsonArray.class::cast).flatMap(JsonArray::stream)
```
halber, line 192: I don't see why we need to convert the result to a JsonArray.
```suggestion
                                .map(Future::result).flatMap(JsonArray::stream)``

halber (changes requested):
All good, just an unnecessary cast that can be removed.

s4heid:
Good catch!

halber (approved):
LGTM

PR #189, opened 2022-10-10
fix: make SystemClassLoader available to Infinispan

pk-work:
When Infinispan is initialized it is using the ContextClassLoader of the current Thread to resolve resources. The ContextClassLoader of a Vert.x event loop is initially null, which causes Infinispan to fail.

### DIFF
--- a/src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
+++ b/src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
@@ -3,6 +3,7 @@ package io.neonbee.cluster;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static io.vertx.core.Future.failedFuture;
 import static io.vertx.core.Future.succeededFuture;
+import static java.lang.ClassLoader.getSystemClassLoader;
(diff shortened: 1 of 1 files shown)


halber (commented):
Looks OK, but is it possible to write a test for it?

pk-work:
@halber I tested several things as discussed yesterday evening. But They didn't worked during Test time Infinispan is initialized on a Thread "Test Runner" not on an EventLoop.

Even with this code, which set up a complete test extension unrelated Vert.x and NeonBee, I can't reproduce it. Maybe we can reproduce it with the "NeonBeeTestRunner" which is maybe introduced by @mkhazanovych.

```java
        NeonBeeOptions opts = new NeonBeeOptions.Mutable().setClustered(true).setClusterManager(INFINISPAN_FACTORY)
                .setIgnoreClassPath(true);
        NeonBee.create(opts).onComplete(
                testContext.succeeding(neonBee -> neonBee.getVertx().close(testContext.succeedingThenComplete())));
```

halber (approved):
LGTM

Issue #199, opened 2022-10-19
NPE in EntityModelLoader when entity model is invalid


s4heid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Given an invalid model, the EntityModelLoader does not properly validate the model. Instead of propagating the error and failing the deployment of the model, a NullPointerException is thrown, which is not catched.

The problematic code is in [`EntityModelLoader.buildModelMap`](https://github.com/SAP/neonbee/blob/cd6ccd993ae1310937f0c85162b191325f825b1f/src/main/java/io/neonbee/entity/EntityModelLoader.java#L204-L206), because `serviceMetaData.getEdm().getEntityContainer()` might be null.

### Expected Behavior

The deployment of the invalid model should fail with an error message, describing the issue.

[Example code](https://github.com/SAP/neonbee/commit/fe7b486585ab995f3118433a1531db089bbd224e), which futurizes `buildModelMap` and handles the error:

```
20:00:16.888 [vert.x-eventloop-thread-10] ERROR io.neonbee.internal.deploy.PendingDeployment - Deployment of Models(models/data-model.csn$models/Esrc.Service.edmx) failed
io.vertx.core.impl.NoStackTraceThrowable: Cannot build model map. Entity Container is empty.
20:00:16.903 [vert.x-eventloop-thread-6] INFO io.neonbee.internal.verticle.ServerVerticle - HTTP server started on port 8080
20:00:16.904 [vert.x-eventloop-thread-10] ERROR io.neonbee.internal.deploy.PendingDeployment - Deployment of Deployables(module:2.0.0-SNAPSHOT) failed
io.vertx.core.impl.NoStackTraceThrowable: Cannot build model map. Entity Container is empty.
20:00:16.904 [vert.x-eventloop-thread-10] ERROR io.neonbee.internal.verticle.DeployerVerticle - Unexpected error occurred during deployment of module from JAR file: /Users/d061008/workspace/neonbee/working_dir/modules/dummy-2.0.0-SNAPSHOT-models.jar
io.vertx.core.impl.NoStackTraceThrowable: Cannot build model map. Entity Container is empty.
```

### Steps To Reproduce

Clone neonbee from the `main` branch, create a folder `working_dir/modules`, and create an invalid cds / edmx file. Start neonbee with `./gradlew run` and you should be able to reproduce the error.

### Environment

```markdown
- OS: Darwin / 21.6.0 / arm64
[...]

PR #201, opened 2022-10-20
fix: #199, add tracing to EntityModelLoader and improve logging

kristian:


src/main/java/io/neonbee/entity/EntityModelLoader.java
```diff
@@ -187,24 +209,43 @@ Future<Void> loadModel(Path csnFile) {
...
+                    .map(Path::toString).map(path -> {
+                        // we do not know if the path uses windows / unix path separators, try both!
+                        return FileSystemHelper.getPathFromMap(associatedModels, path);
+                    }).map(this::parseEdmxModel).collect(Collectors.toList())).onComplete(result -> {
+                        if (LOGGER.isTraceEnabled()) {
+                            LOGGER.trace("Parsing associated models of {} {}", csnFile,
+                                    result.succeeded() ? "succeeded" : "failed");
+                        }
+                    }).onSuccess(compositeFuture -> {
+                        buildModelMap(cdsModel, compositeFuture.<ServiceMetadata>list());
```
s4heid, line 229: In my opinion, this does not properly handle the error. `buildModelMap` might throw a `RuntimeException`, when building the model map is not possible. In that case I think it would be better to let `parseModel` return a `failedFuture`.

Same with `EntityModelManager.loadModel`.

s4heid (changes requested):
Thanks @kristian for providing this fix! 🔥 

I tested it locally with some dummy models and the additional log output is very helpful in understanding what's actually going on. As we now have so many logs (on `TRACE` level) it could be a bit difficult to follow/read them if you have many (let's say more than 10) models. Thought about adding an identifier for the model loading that can be put as correlation id. If you think that's useful we can also do it in a separate change in order to keep this one comprehensible, as this would be more like an enhancement than actually fixing something.

Apart from that I only have this single concern, which I added as a review comment.

kristian:
I agree that having many logs could be a bit confusing, however as they are TRACE logs I don't think they have to be very readable, because they are for the lowest tier of root cause analysis. So yes, if we go for something like a model identifier for the logging, please let's do this in another PR. For your comment, I will answer it separately.

src/main/java/io/neonbee/entity/EntityModelLoader.java
kristian: Mhm I agree generally with your statement, that throwing a plain `RuntimeException` is most strange. However I don't think it is required to return a `failedFuture` either... The question is: Is a CDS model w/o a service (in that case `getNamespace` throws the runtime exception) a failure? Well according to the CDS compiler yes, you cannot compile a CDS model w/o a service to EDMX/CSN, however their Java API (`CdsModelBuilder`) allows to create a model w/o a service (I created a test for it). Thus I think the right approach would be: We just ignore models w/o services for now and, like the CDS compiler on empty services, log a warning. That is I think the most "gentle" approach to this, what do you think?

kristian:
I made a suggestion, please have a look.

src/main/java/io/neonbee/entity/EntityModelLoader.java
s4heid: > The question is: Is a CDS model w/o a service (in that case getNamespace throws the runtime exception) a failure?

I would say yes, just because it takes much longer for a developer to find out why the endpoint is not showing data if we only write a log (on `TRACE` level which is anyways hard to find - as we write many logs) rather than failing the deployment. I don't think anyone would create such models if it wasn't a mistake.

On the other hand, I'm with you on the argument about the `CdsModelBuilder`, but wouldn't necessarily see that as a requirement that we accept such models.

kristian: Would you mind taking a look at my current revision of this change, I think the way I now logged it as a WARNING is the right way to go, if you also agree, I would fix the last issues of the voter and merge.

s4heid (approved):

PR #202, opened 2022-10-20
ci: fix tests

kristian:
- Re-enable parallel test execution (old issues are fixed)
- Better to reset the FakeClusterManager more often than not
- The Infinispan cluster manager must be stopped after tests
- A test with a timeout of over a hour is not good
- Tests that create own Vert.x instances or NeonBee(s) must take care to also shut them down properly, which can be very tricky. Generally it is recommended to use existing test bases or extensios, as they take care to properly shut down the instance(s).
- Tests that use the FakeClusterManager must run isolated
- Test methods that are in a NeonBeeTestBase, but not require NeonBee, can now be tagged with a @tag(DOESNT_REQUIRE_NEONBEE)
- Improve StaleThreadChecker to locate all tests creating stale threads (set VERBOSE = true, disable parallel execution and enable showStandardStreams in build.gradle to run analysis)

### DIFF
--- a/src/test/java/io/neonbee/cluster/LocalRequestClusterTest.java
+++ b/src/test/java/io/neonbee/cluster/LocalRequestClusterTest.java
@@ -8,6 +8,7 @@ import java.util.concurrent.TimeUnit;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.parallel.Isolated;
 
 import io.neonbee.NeonBee;
 import io.neonbee.NeonBeeExtension;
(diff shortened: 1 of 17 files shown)


src/test/java/io/neonbee/cluster/LocalRequestClusterTest.java
```diff
@@ -26,6 +27,7 @@
 import io.vertx.junit5.VertxTestContext;
 
 @ExtendWith(NeonBeeExtension.class)
+@Isolated("Some of the methods in this test class run clustered and use the FakeClusterManager for it. The FakeClusterManager uses a static state and can therefore not be run with other clustered tests.")
```
pk-work, line 30: Is it possible to add the @Isolated to all tests that are based on NeonBeeExtension?

pk-work (commented):
Thanks @kristian for your great work and bringing back the parallel execution!!!

I only had one comment, I hope we can add the isolated tag somehow to NeonBeeExtension maybe with a Junit rule or so.

src/test/java/io/neonbee/cluster/LocalRequestClusterTest.java
kristian: Thanks for the review @pk-work. I have checked the JUnit documentation and also some of their code, but I found no apparent way, how that could be feasible unfortunately... good idea though.

pk-work (approved):

PR #203, opened 2022-10-21
fix: odata query encoding

halber:
The OdataRequest RawQueryPath is expected to be in Percent-encoding, but it was decoded.


kristian (approved):
lgtm

PR #204, opened 2022-10-23
feat: introduce a NeonBeeExtension.TestBase / fix: close Infinispan cache container after NeonBee shutdown

kristian:


kristian:
I would like to add one more thing I realized about the extension vs. testbase approach to this PR, so please do not merge yet. Thanks

src/test/java/io/neonbee/test/base/ClusteredNeonBeeTestBase.java
```diff
@@ -0,0 +1,24 @@
...
+ * The {@link ClusteredNeonBeeTestBase} is a little syntax sugar for test classes, that are dealing with clustered
+ * tests. This class essentially wraps two things together and thusly decreases the risk of making a mistake:
+ * <ol>
+ * <li>Cluster tests should always run isolated. Thus the {@link ClusteredNeonBeeTestBase} is {@link Isolated}.</li>
+ * <li>This test base overrides the {@link #adaptOptions(TestInfo, Mutable)} method, in order to set all tests to run
+ * clustered by default.</li>
+ * </ol>
+ */
+@Isolated("Clustered NeonBee tests must always run isolated. The default FakeClusterManager has a static state and multiple real cluster manager should never run side-by-side for multiple tests to avoid interfering")
+public class ClusteredNeonBeeTestBase extends NeonBeeTestBase {
```
pk-work, line 18: I don'T have a good feeling for ClusteredNeonBeeTestBase, because NeonBeeTestBase can only handle one Node .. so you will NEVER create a cluster with NeonBeeTestBase and this is what "ClusteredNeonBeeTestBase" looks like.

I would suggest to have a method like "provideUserPrincipal" to "enable" the cluster mode for this single NeonBee instance.

I then would rename ClusteredNeonBeeExtensionTestBase into ClusteredNeonBeeTestBase.

kristian: You are right. I adapted the change as you suggested but renamed the new test base to ClusterExtensionTestBase to make it more distingushed from the NeonBeeTestBase

pk-work (approved):

pk-work:
Good Job!

PR #208, opened 2022-10-25
ci: add dependabot configuration

s4heid:


pk-work (dismissed):

pk-work (dismissed):

pk-work (approved):
Let's see what we will break :)

PR #210, opened 2022-10-28
fix: odata query encoding and DataQuery setQuery

halber:
In this pull request two problems are fixed

1.  DataQuery cannot correctly process query strings
    The DataQuery expects a decoded query string in the setQuery method and
    in the constructors that accept a query parameter. These methods are
    now deprecated since this cannot be fixed. A new method setRawQuery is
    introduced which accepts the encoded query string and creates the
    parameters from it.

3. odata query encoding in EntityVerticle
    The parseUri method expects the query parameter to be in percent
    encoding, but it was decoded.

### DIFF
--- a/src/main/java/io/neonbee/data/DataQuery.java
+++ b/src/main/java/io/neonbee/data/DataQuery.java
@@ -6,6 +6,9 @@ import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.mapping;
 import static java.util.stream.Collectors.toCollection;
 
+import java.net.URLDecoder;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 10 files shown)


pk-work:
Maybe this is the reason why NeonBee was not able to handle single quotes in the past. Please have a look at https://github.com/SAP/neonbee/commit/4a0665a799205d557c0387cbb06c77a84442f420

This commit contains tests which have failed. Can you add these tests to your change (if you not already have similar ones) to check if this issue is now solved?

halber:
> Maybe this is the reason why NeonBee was not able to handle single quotes in the past. Please have a look at [4a0665a](https://github.com/SAP/neonbee/commit/4a0665a799205d557c0387cbb06c77a84442f420)
> 
> This commit contains tests which have failed. Can you add these tests to your change (if you not already have similar ones) to check if this issue is now solved?

I added the test case, but I had to change it a bit to make the test green.

BenniR6 (approved):

BenniR6 (approved):
Unfortunately a lot of deprecations needed. 
I can not see a better idea atm.

PR #211, opened 2022-10-31
fix #188: HTTP Header fields should be case-insensitive

### LINKED ISSUE #188: HTTP Header fields should be case-insensitive

### Current Behavior

The getters of HTTP headers in `DataQuery` do not handle the names case-insensitively. However, they should be; as stated by the HTTP RFC 2616 [Section 4.2](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):

> Field names are case-insensitive.

### Expected Behavior

Field names are handled case-insensitively.

### Steps To Reproduce

_No response_

### Environment

_No response_

### Relevant log output

_No response_

### Anything else?

_No response_

livkovacova:
- headers field is now CaseInsensitiveMap instead of Map
- added tests for change coverage

### DIFF
--- a/src/main/java/io/neonbee/data/DataQuery.java
+++ b/src/main/java/io/neonbee/data/DataQuery.java
@@ -14,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.TreeMap;
 import java.util.function.Function;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -54,7 +55,7 @@ public final class DataQuery { // NOPMD not a "god class"
 
     @VisibleForTesting
     @JsonProperty
-    Map<String, List<String>> headers;
+    Map<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
 
     @VisibleForTesting
     @JsonSerialize(using = BufferSerializer.class)
@@ -186,7 +187,7 @@ public final class DataQuery { // NOPMD not a "god class"
     public DataQuery(DataAction action, String uriPath, String query, Map<String, List<String>> headers, Buffer body) {
         this.action = action;
         this.setUriPath(uriPath); // Calling the setter here, because there is an additional check implemented.
-        this.headers = CollectionHelper.mutableCopyOf(headers != null ? headers : Map.of());
+        this.headers = CollectionHelper.mapToCaseInsensitiveTreeMap(headers != null ? headers : Map.of());
         this.parameters = parseQueryString(query);
         this.body = CollectionHelper.copyOf(body);
     }
@@ -385,7 +386,7 @@ public final class DataQuery { // NOPMD not a "god class"
      * @return the DataQuery for chaining
      */
     public DataQuery setHeaders(Map<String, List<String>> headers) {
-        this.headers = CollectionHelper.mutableCopyOf(headers);
+        this.headers = CollectionHelper.mapToCaseInsensitiveTreeMap(headers);
         return this;
     }
 
--- a/src/main/java/io/neonbee/internal/helper/CollectionHelper.java
+++ b/src/main/java/io/neonbee/internal/helper/CollectionHelper.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -12,6 +13,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.function.BiFunction;
 import java.util.function.Supplier;
 import java.util.stream.Collector;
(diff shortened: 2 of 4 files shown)


src/main/java/io/neonbee/data/DataQuery.java
```diff
@@ -54,7 +56,7 @@ public final class DataQuery { // NOPMD not a "god class"
 
     @VisibleForTesting
     @JsonProperty
-    Map<String, List<String>> headers;
+    Map<String, List<String>> headers = new CaseInsensitiveMap<>();
```
halber, line 59:
```suggestion
    Map<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
Then we would not need the dependency on commons-collections4.

src/main/java/io/neonbee/data/DataQuery.java
```diff
@@ -186,7 +188,9 @@ public DataQuery(DataAction action, String uriPath, String query, Map<String, Li
     public DataQuery(DataAction action, String uriPath, String query, Map<String, List<String>> headers, Buffer body) {
         this.action = action;
         this.setUriPath(uriPath); // Calling the setter here, because there is an additional check implemented.
-        this.headers = CollectionHelper.mutableCopyOf(headers != null ? headers : Map.of());
+        if (headers != null) {
+            this.headers.putAll(headers);
+        }
```
halber, line 193: You create only a shallow copy of the header map. This breaks the copy method. This way, you could create a copy using the copy method and change the values of the two DataQuery objects by changing them in one object.

Example test case:
```
DataQuery query1 = new DataQuery("uri", "name=Hodor", Map.of("header1", new ArrayList<>(List.of("value1"))));
assertThat(query1.getHeaderValues("header1")).hasSize(1);
DataQuery query2 = query1.copy();
query2.getHeaderValues("header1").add("value2");
assertThat(query1.getHeaderValues("header1")).hasSize(1);
```

halber (changes requested):

halber:
Hi @livkovacova,
do you have any questions about my comments, or do you just not have time to work on this right now? Just leave a message so we know that this PR is still work in progres.

livkovacova:
Hello, sure, this PR is still work in progress, I will commit changes according to recommendations as soon as possible, thank you for your comments, if I have more questions I ask you, so that we are both satisfied

src/main/java/io/neonbee/internal/helper/CollectionHelper.java
```diff
@@ -138,6 +139,21 @@ public static <T> T copyOf(T object) {
...
+     * keys and values and adding them to a new case-insensitive treemap {@link TreeMap}.
+     *
+     * @param <K> the key type
+     * @param <V> the value type
+     * @param map the map to copy
+     * @return a new case-insensitive treemap as mutable deep copy of the given map
+     */
+    public static <K extends String, V> Map<K, V> mapToCaseInsensitiveTreeMap(Map<K, V> map) {
+        return Optional.ofNullable(map).map(Map::entrySet).map(Set::stream).orElseGet(Stream::empty)
+                .collect(Collectors.toMap(entry -> copyOf(entry.getKey()), entry -> copyOf(entry.getValue()),
```
halber, line 153: With null values we would currently run into an NPE because Collectors.toMap calls merge and the TreeMap does not override the default method of java.util.Map#merge.
For example see io.neonbee.internal.helper.CollectionHelper.NullLiberalMergingHashMap

halber: Unfortunately, the test for this is missing. Could you add this test to your PR?
```
package io.neonbee.internal.helper;

import static com.google.common.truth.Truth.assertThat;

import java.util.HashMap;
import java.util.Map;
import java.util.TreeMap;

import org.junit.jupiter.api.DisplayName;
import org.junit.jupiter.api.Test;

class CollectionHelperTest {

    @Test
    @DisplayName("Copy map with mutableCopyOf")
    void testMutableCopyOf() {
        HashMap<String, Object> map = new HashMap<>();
        map.put("k", null);
        map.put("key", "value");
        map.put("k1", 1);
        Map<String, Object> resultMap = CollectionHelper.mutableCopyOf(map);
        assertThat(resultMap).isInstanceOf(HashMap.class);
        assertThat(resultMap).containsExactlyEntriesIn(map);

    }

    @Test
    @DisplayName("Copy map with mapToCaseInsensitiveTreeMap")
    void testMapToCaseInsensitiveTreeMap() {
        HashMap<String, Object> map = new HashMap<>();
        map.put("k", null);
        map.put("key", "value");
        map.put("k1", 1);
        Map<String, Object> resultMap = CollectionHelper.mapToCaseInsensitiveTreeMap(map);
        assertThat(resultMap).isInstanceOf(TreeMap.class);
        assertThat(resultMap).containsExactlyEntriesIn(map);
    }
}
```

halber (changes requested):
Overall a good PR! Currently the Lint Commit Messages action is failing and there is one small issue in the code.
Could you also squash your commits into a single commit?
Thank you very much!

src/main/java/io/neonbee/internal/helper/CollectionHelper.java
livkovacova: Thank you for review, I've made requested changes.

halber (approved):
LGTM

s4heid:
@livkovacova thank you for the contribution! this change looks really good now 🚀

PR #212, opened 2022-11-04
Include version file inside jar under META-INF/neonbee

### LINKED ISSUE #198: Versioning information of the NeonBee jar

### The Problem

At the moment there is no (easy) way to determine the NeonBee release version from the jar.

### Desired Solution

Other projects like Vert.x include a file (e.g. `vertx-version.txt`) which contains the release version into the `META-INF` directory. This would solve the problem.

### Alternative Solutions

_No response_

### Additional Context

In addition to the file based solution, Vert.x offers a [CLI command](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/impl/launcher/commands/VersionCommand.java). This could be a nice to have extra, but the file based approach is sufficient for the beginning.

wowselim:
This PR adds a gradle task that generates a version file under `META-INF/neonbee` called `neonbee-version.txt`.

closes #198

### DIFF
--- a/gradle/distribution.gradle
+++ b/gradle/distribution.gradle
@@ -51,6 +51,18 @@ shadowJar {
     mergeServiceFiles()
 }
 
+task generateVersionFile {
+    doLast {
+        def resourceDir = sourceSets.main.output.resourcesDir
+        def versionFile = new File(resourceDir, "META-INF/neonbee/neonbee-version.txt")
+        versionFile.text = project.version
+    }
+}
+
+tasks.withType(Jar) {
+    dependsOn generateVersionFile
+}
+
 /** Builds the sources JAR */
 task sourcesJar(type: Jar) {
     from sourceSets.main.allJava


gradle/distribution.gradle
```diff
@@ -51,6 +51,18 @@ shadowJar {
     mergeServiceFiles()
 }
 
+task generateVersionFile {
+    doLast {
+        def resourceDir = sourceSets.main.output.resourcesDir
+        def versionFile = new File(resourceDir, "META-INF/neonbee/neonbee-version.txt")
+        versionFile.text = project.version.replace('-SNAPSHOT', '')
```
pk-work, line 58: Could it be beneficial if we don't loose the snapshot information? @s4heid What do you think?

pk-work (changes requested):
The code looks good, but the PR isn't following the NeonBee development concept [1]. Please have a look into the review chapter [2].

I'd really appreciate if you could squash the commits and write a valid commit message. But if you don't have time, I can do it also for you.

[1] https://github.com/SAP/neonbee/blob/main/docs/development_concept.md
[2] https://github.com/SAP/neonbee/blob/main/docs/development_concept.md#review

pk-work:
@wowselim Thank you very much for you contribution!

gradle/distribution.gradle
s4heid: I think the `-SNAPSHOT` version is not very informative, because snapshot releases are not definite. It would be rather interesting to see the git commit hash, instead.

At the end you are right, @pk-work, [snapshot releases](https://github.com/SAP/neonbee/blob/4f0bfaeabeb58f5efc8ba2c0829642b7dd6a99fa/.github/workflows/publish.yml#L240) are published are published to maven central and it would be weird if they contain a non snapshot version. So, better keep the suffix.

pk-work: Okay, then lets keep the suffix. @wowselim is this also fine for you?

wowselim: done!

pk-work (approved):

gradle/distribution.gradle
pk-work: Thank you very much for your contribution and your patience.

PR #214, opened 2022-11-14
feat: load Hazelcast configuration from file

halber:
Add loading hazelcast configurations from a file outside the classpath. First we search for the file on the classpath. If the file cannot be found, we try to load the file with FileSystemXmlConfig.

### DIFF
--- a/src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
+++ b/src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
@@ -5,12 +5,16 @@ import static io.vertx.core.Future.failedFuture;
 import static io.vertx.core.Future.succeededFuture;
 import static java.lang.ClassLoader.getSystemClassLoader;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
 
 import org.infinispan.manager.DefaultCacheManager;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.hazelcast.config.ClasspathXmlConfig;
+import com.hazelcast.config.Config;
+import com.hazelcast.config.FileSystemXmlConfig;
 
 import io.neonbee.NeonBeeOptions;
 import io.vertx.core.Future;
@@ -23,6 +27,7 @@ public abstract class ClusterManagerFactory {
      * The ClusterManagerFactory for Hazelcast.
      */
     public static final ClusterManagerFactory HAZELCAST_FACTORY = new ClusterManagerFactory() {
+
         @Override
         protected String getDefaultConfig() {
             return "hazelcast-cf.xml";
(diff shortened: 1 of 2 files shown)


src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
```diff
@@ -31,8 +36,35 @@ protected String getDefaultConfig() {
...
+
+            return Future.future(promise -> {
+                Config config;
+                if (exitsInClasspath(effectiveConfig)) {
+                    config = new ClasspathXmlConfig(getSystemClassLoader(), effectiveConfig, System.getProperties());
+                } else {
+                    try {
+                        config = new FileSystemXmlConfig(effectiveConfig, System.getProperties());
+                    } catch (FileNotFoundException e) {
+                        throw new RuntimeException(e);
```
pk-work, line 48: Would it make sense not to throw an exception here, but to return a failed future describing that the configuration was not found in both the classpath and the filesystem?

src/test/java/io/neonbee/cluster/ClusterManagerFactoryTest.java
```diff
@@ -84,4 +92,31 @@ void testDefaultFactories(ClusterManagerFactory cmf, String defaultConfig, Class
...
+    static Stream<Arguments> withHazelcastConfig() throws IOException {
+        File hazelcastTestFile = FileSystemHelper.createTempDirectory().resolve("test-hazelcast.xml").toFile();
+        File file = TEST_RESOURCES.resolve("hazelcast-localtcp.xml").toFile();
+        FileUtils.copyFile(file, hazelcastTestFile);
+
+        Mutable options = new Mutable();
+        options.setClustered(true).setClusterConfig(hazelcastTestFile.getPath());
+
+        Arguments hazelcast = Arguments.of(hazelcastTestFile.getPath(), FileSystemXmlConfig.class, options);
+        Arguments infinispan = Arguments.of("hazelcast-cf.xml", ClasspathXmlConfig.class, new Mutable());
```
pk-work, line 105: I think infinispan is a copy paste mistake

pk-work (changes requested):

src/main/java/io/neonbee/cluster/ClusterManagerFactory.java
halber: I wrapt the Exception in a RuntimeException to fail the promise. I could add an additional descriptive message to the RuntimeException.
I also could fail the promise at this point to avoid warping the FileNotFoundException in a RuntimeException. 
```
try {
    config = new FileSystemXmlConfig(effectiveConfig, System.getProperties());
} catch (FileNotFoundException e) {
    promise.fail(e);
    return;
}
```
At the moment, I don't see any real advantage of one method over the other.

pk-work: You are right the promise is catching the exception

pk-work (approved):

PR #216, opened 2022-11-16
refactor: extract registry logic

halber:
There is a registry logic that is commonly used. In order to reuse this logic, I have moved this logic to a new WriteSafeRegistry class.

### DIFF
--- a/src/main/java/io/neonbee/entity/EntityVerticle.java
+++ b/src/main/java/io/neonbee/entity/EntityVerticle.java
@@ -26,7 +26,8 @@ import io.neonbee.data.DataContext;
 import io.neonbee.data.DataQuery;
 import io.neonbee.data.DataRequest;
 import io.neonbee.data.DataVerticle;
-import io.neonbee.internal.SharedDataAccessor;
+import io.neonbee.internal.Registry;
+import io.neonbee.internal.WriteSafeRegistry;
 import io.neonbee.internal.helper.AsyncHelper;
 import io.neonbee.internal.verticle.ConsolidationVerticle;
 import io.neonbee.logging.LoggingFacade;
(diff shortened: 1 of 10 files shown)


src/main/java/io/neonbee/entity/EntityVerticle.java
```diff
@@ -250,34 +255,19 @@ private Future<Void> announceEntityVerticle(Vertx vertx) {
...
-                                            });
-                                        });
+                                String qualifiedName = getQualifiedName();
+                                return getRegistry(vertx).register(name, qualifiedName);
                             }).collect(Collectors.toList());
                     return allComposite(announceFutures).mapEmpty();
                 });
     }
+
+    static WriteSafeRegistry getRegistry(Vertx vertx) {
```
s4heid, line 267: You ensure thread safety via the locking in the `WriteSafeRegistry` itself, but honestly I don't feel comfortable with this singleton approach. It comes with several other drawbacks like the fact that we need to introduce a `EntityVerticleExtension`. I would even rather manage it centrally via the NeonBee instance instead. Did you consider this?

src/main/java/io/neonbee/internal/WriteSafeRegistry.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    /**
+     * Unregister the value in the {@link NeonBee} async shared map from the sharedMapKey.
+     *
+     * @param sharedMapKey the shared map key
+     * @param value        the value to register
+     * @return the future
+     */
+    public Future<Void> unregister(String sharedMapKey, String value) {
```
s4heid, line 116: this method is completely untested. Also noticed there is no `WriteSafeRegistryTest` class; might be a good idea to introduce and write a test for `unregister`.

src/main/java/io/neonbee/entity/EntityVerticle.java
s4heid: Maybe, let's have a short sync after standup to discuss available options.

s4heid (changes requested):

pk-work:
Also from my side a big thank you @halber !

PR #218, opened 2022-11-24
feat: remove MetricsVerticle

pk-work:
Obviously the MetricsVerticle only exists to log continuously metrics if TRACE is activated. This is a very special purpose and shouldn't be part of the system, verticles. Also, the name MetricsVerticle potentially raises false expectations.


halber (approved):
I can't find any reason why it wouldn't be ok to delete this.

PR #222, opened 2022-12-06
fix: getName for dummy EntityVerticle

halber:
The EntityVerticle#getName returns the same name for each EntityVerticle created with the createDummyEntityVerticle method. Therefore, all created EntityVerticles are listing on the same event bus address and Vert.x is calling the EntityVerticles in a round-robin fashion.

To fix this, the getName method for the created EntityVerticle objects is overridden with Mockito#spy.

### DIFF
--- a/src/test/java/io/neonbee/test/helper/DummyVerticleHelper.java
+++ b/src/test/java/io/neonbee/test/helper/DummyVerticleHelper.java
@@ -1,6 +1,8 @@
 package io.neonbee.test.helper;
 
 import static io.vertx.core.Future.succeededFuture;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
 
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


src/test/java/io/neonbee/test/helper/DummyVerticleHelper.java
```diff
@@ -249,6 +251,10 @@ public Future<Set<FullQualifiedName>> entityTypeNames() {
                     return succeededFuture(Set.of(fqn));
                 }
             };
+
+            EntityVerticle spy = spy(ev);
+            when(spy.getName()).thenReturn(String.format("_%s-%d", fqn, ev.getClass().getName().hashCode()));
```
pk-work, line 256: I'm not sure if FQN is a real improvement. In case I want to deploy 2 different DummyEntityVerticles that provide the same Entity, you would end up in the same situation. It's better than now, because now you can only deploy ONE EntityVerticle, but no real solution, isn't it?

kristian:
For tests the getName should not really matter, thus my suggestion would be to simply use a the hashCode of the newly created object itself which should be always unique. E.g.:

"_dummyEntityVerticle"+ev.hashCode()

src/test/java/io/neonbee/test/helper/DummyVerticleHelper.java
halber: I have updated the code to use the object identity hash code instead of the class name hash code.

pk-work (approved):

PR #228, opened 2023-01-11
docs: replace getting started section with link to neonbee-examples

pk-work:


s4heid:
If we plan to write additional user documentation about Data Verticles / Entity Verticles, the deleted section might be a good starting point - just something we should keep in mind.

s4heid (approved):

PR #224, opened 2022-12-23
build(deps): upgrade vert.x to 4.3.7

ada89:

PR #229, opened 2023-01-17
feat(health): add status endpoint & support filtering

s4heid:
With this change a `StatusEndpoint` is introduced, which allows to display health onformation in a non-verbose way. This endpoint is meant to be public and usable for external system monitoring tools. The endpoint is mounted at /status and is part of the default endpoints of NeonBee. The /health endpoint will continue to serve health data in a verbose way, and it should be considered by the user to protect this endpoint as it might expose system information that should not be available to external parties.

In addition, the `HealthCheckHandler` is extended to support health check based filtering vie the request path of the /health endpoint.

### DIFF
--- a/src/main/java/io/neonbee/endpoint/health/HealthCheckHandler.java
+++ b/src/main/java/io/neonbee/endpoint/health/HealthCheckHandler.java
@@ -1,64 +1,171 @@
 package io.neonbee.endpoint.health;
 
+import static io.neonbee.internal.helper.StringHelper.EMPTY;
 import static java.util.Objects.requireNonNull;
 
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
 import org.apache.olingo.commons.api.http.HttpStatusCode;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 
 import io.neonbee.data.internal.DataContextImpl;
 import io.neonbee.health.HealthCheckRegistry;
 import io.neonbee.logging.LoggingFacade;
 import io.netty.handler.codec.http.HttpResponseStatus;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.http.HttpHeaders;
 import io.vertx.core.http.HttpServerResponse;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.healthchecks.CheckResult;
+import io.vertx.ext.web.Route;
 import io.vertx.ext.web.RoutingContext;
 
+/**
+ * {@link HealthCheckHandler} is a class that implements the Vert.x {@link RoutingContext} Handler interface. It is used
+ * to handle health check requests by providing a RESTful API for the health status.
+ *
+ * The class collects the health check results from a central {@link HealthCheckRegistry} and formats the response with
+ * appropriate status codes based on the results. This can be retrieved for all checks, or, if a specific and existing
+ * health check name is passed in the request path, only health checks for the single health check are returned. By
+ * default, the HTTP response is verbose and provide detailed information about every single health check. If the
+ * verbose flag is set to false via the constructor, only the overall status and a version number is returned.
+ */
 public class HealthCheckHandler implements Handler<RoutingContext> {
     @VisibleForTesting
     static final String APPLICATION_JSON_CHARSET_UTF_8 = "application/json;charset=UTF-8";
+
+    @VisibleForTesting
+    static final String VERSION_FILE = "META-INF/neonbee/neonbee-version.txt";
+
+    private static final String DEV_VERSION = "0.0.0+dev";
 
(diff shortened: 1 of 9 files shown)


s4heid:
Test failures are related to weak performance of the voter worker. Voter should be green after #224 is merged.

src/main/java/io/neonbee/endpoint/health/HealthCheckHandler.java
```diff
@@ -1,64 +1,168 @@
...
+    /**
+     * Parses the health check name from the request path if provided.
+     *
+     * @param rc the {@link RoutingContext} with the request
+     * @return the health check name if provided, empty String otherwise
+     */
+    private String parseCheck(RoutingContext rc) {
+        String requestPath = rc.request().path();
+        String routePath =
+                Optional.ofNullable(rc.currentRoute()).map(Route::getPath).orElse(EMPTY).replaceAll("/+$", EMPTY) + "/";
```
halber, line 101:
```suggestion
        String routePath = Optional.ofNullable(rc.currentRoute())
                .map(Route::getPath)
                .map(path -> path.replaceAll("/+$", EMPTY) + "/")
                .orElse("/");
```
This way we execute replaceAll only on a non null path.

src/main/java/io/neonbee/health/HealthCheckRegistry.java
```diff
@@ -217,31 +234,40 @@ private List<Future<JsonArray>> sendDataRequests(JsonArray qualifiedNames, DataC
...
+                            // requests an external service which requires credentials, but the credentials are not
+                            // available on the node due to misconfiguration. In this scenario, keep the existing entry.
+                        }
+                    } else {
+                        consolidatedChecks.put(checkId, checkResult);
+                        if (DOWN.equals(status)) {
+                            aggregatedStatus.set(DOWN);
+                        }
+                    }
+                });
```
halber, line 266: I see no advantage in using a predicate here. I find it rather an unusual solution. In my opinion, it would be clearer to change the predicate to a method. I would also change the disjunction to two filter calls.

```
suggestion

        healthCheckResults.stream()
                .filter(checkResult -> Strings.isNullOrEmpty(check))
                .filter(checkResult -> isIncluded(check, checkResult))
                .filter(checkResult -> checkResult.getString(ID_KEY) != null
                        && checkResult.getString(STATUS_KEY) != null)
                .forEach(checkResult -> {
                    String checkId = checkResult.getString(ID_KEY);
                    String status = checkResult.getString(STATUS_KEY);

                    if (consolidatedChecks.containsKey(checkId)) {
                        if (!status.equals(consolidatedChecks.get(checkId).getString(STATUS_KEY))) {
                            LOGGER.correlateWith(dataContext).warn("Detected inconsistent status of health check {}",
                                    checkId);
                            // some nodes report a different status of the check than other nodes. This might be due to
                            // some inconsistent uptime of the service, but more likely that some requirements to
                            // perform a successful health check are not met by some nodes. Example: the health check
                            // requests an external service which requires credentials, but the credentials are not
                            // available on the node due to misconfiguration. In this scenario, keep the existing entry.
                        }
                    } else {
                        consolidatedChecks.put(checkId, checkResult);
                        if (DOWN.equals(status)) {
                            aggregatedStatus.set(DOWN);
                        }
                    }
                });

        if (LOGGER.isTraceEnabled()) {
            LOGGER.correlateWith(dataContext).trace("Consolidated health checks: " + consolidatedChecks);
        }

        return new JsonObject().put(CHECKS_KEY, new JsonArray(new ArrayList<>(consolidatedChecks.values())))
                .put(STATUS_KEY, aggregatedStatus.get()).put(OUTCOME_KEY, aggregatedStatus.get());
    }

    private boolean isIncluded(String check, JsonObject checkResult) {
        String c = checkResult.getString(ID_KEY);
        return c.equals(check) || (c.startsWith("node") && c.endsWith(check));
    }
```

halber (changes requested):

src/main/java/io/neonbee/endpoint/health/HealthCheckHandler.java
s4heid: that's better!

src/main/java/io/neonbee/health/HealthCheckRegistry.java
s4heid: extracted it into a private method. changing the filter as suggested is not equivalent

src/main/java/io/neonbee/endpoint/health/HealthCheckHandler.java
```diff
@@ -1,64 +1,168 @@
...
+    /**
+     * Sets the status code of the response object based on the value of "status" key in the input JsonObject. It also
+     * adds the response header and based on verbose mode sets a detailed or non-detailed response body.
+     *
+     * @param rc      the current {@link RoutingContext}
+     * @param json    all consolidated health checks
+     * @param verbose if response should be verbose or not
+     * @return a succeeded Future if handling the requests is successful, a failed Future otherwise.
+     */
+    private Future<Void> handleRequest(RoutingContext rc, JsonObject json, boolean verbose) {
```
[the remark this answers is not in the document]
s4heid: should not be an issue due to the return

halber (approved):
LGTM

PR #249, opened 2023-02-01
test(odata): unit tests for OData Batch Request handling

tschuba:
New unit tests covering OData batch request processing


pk-work:
I will have a look

PR #285, opened 2023-03-15
feat: allow to pass count value via response hints

pk-work:
This commit allows to pass the count information via response hint, which is a hugh performance benefit, because it is no longer necessary to send all entities back to the ODataEndpoint.


halber (commented):
Just a hint: If someone would put something other than integer into the hint, a ClassCastException would occur. I think this is problematic because in the forwardRequest method processPromise is not complete or failed.

halber (approved):
LGTM

PR #293, opened 2023-03-23
feat: add CorsHandlerFactory

pk-work:
This change adds a CorsHandlerFactory, that allows to add and configure a CORS handler via the ServerConfig.

### DIFF
--- a/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
+++ b/resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
@@ -100,4 +100,23 @@ config:
     - io.neonbee.internal.handler.factories.TimeoutHandlerFactory
     - io.neonbee.internal.handler.factories.SessionHandlerFactory
     - io.neonbee.internal.handler.factories.CacheControlHandlerFactory
+    - io.neonbee.internal.handler.factories.CorsHandlerFactory
     - io.neonbee.internal.handler.factories.DisallowingFileUploadBodyHandlerFactory
+
+    # configures the CORS handler (Disabled by default)
+    cors:
+        enabled: false # Enables the CORS handler.
+        origins: # Set the list of allowed static origins.
+            - http://foo.bar
+        relativeOrigins: # Set the list of allowed relative origins.
+            - "^http\\://.*.foo.bar$"
+        allowedMethods: # Set a set of allowed methods. Uppercase required!
+            - GET
+            - PUT
+        allowedHeaders: # The allowed HTTP headers.
+            - foo
+            - bar
+        exposedHeaders: # The exposed headers.
+            - foobar
+        maxAgeSeconds: 1337 # Set how long the browser should cache the information.
+        allowCredentials: false # Set whether credentials are allowed or not.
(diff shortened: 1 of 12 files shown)


resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
```diff
@@ -100,4 +100,23 @@ config:
...
+    - io.neonbee.internal.handler.factories.DisallowingFileUploadBodyHandlerFactory
+
+    # configures the CORS handler (Disabled by default)
+    cors:
+        enabled: false # Enables the CORS handler.
+        origins: # Set the list of allowed static origins.
+            - http://foo.bar
+        relativeOrigins: # Set the list of allowed relative origins.
+            - "^http\\://.*.foo.bar$"
+        allowedMethods: # Set a set of allowed methods. Upper case required!
```
gedack, line 113: Typo: "Uppercase" instead of "upper case"

resources/config/io.neonbee.internal.verticle.ServerVerticle.example.yaml
```diff
@@ -100,4 +100,23 @@ config:
...
+            - GET
+            - PUT
+        allowedHeaders: # Set a set of allowed headers.
```
gedack, line 116: "Set a set" in "Set a set of allowed headers" sounds strange. Why not just something like "The allowed HTTP headers".
Similar case above and below.

src/main/java/io/neonbee/config/CorsConfig.java
```diff
@@ -0,0 +1,271 @@
...
+/**
+ * Global CORS configuration.
+ */
+@DataObject(generateConverter = true, publicConverter = false)
+public class CorsConfig {
+    private boolean enabled;
+
+    private List<String> origins;
+
+    private List<String> relativeOrigins;
```
gedack, line 20: What is the behaviour if we add the same origin/relativeOrigin multiple times? Maybe a Set makes more sense?

src/main/java/io/neonbee/config/CorsConfig.java
```diff
@@ -0,0 +1,271 @@
...
+        this.allowCredentials = allowCredentials;
+        return this;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof CorsConfig)) {
```
gedack, line 254: I do not expect subclasses of CorsConfig. So wouldn't  "... getClass() != o.getClass() ... " be the more appropriate approach, because only the exact type is important?

gedack (changes requested):

src/main/java/io/neonbee/config/CorsConfig.java
pk-work: Now the voter fails with:

```
/home/runner/work/neonbee/neonbee/src/main/java/io/neonbee/config/CorsConfig.java:250: warning: [EqualsGetClass] Prefer instanceof to getClass when implementing Object#equals.
```

So I will change it back

src/main/java/io/neonbee/config/CorsConfig.java
gedack: Awesome, thank you for clarification.

gedack (approved):
LGTM

PR #318, opened 2023-05-23
feat: return 201 (not 204) for POST requests to entity verticles and …

dapeng-wang:
…optional Location-header

### DIFF
--- a/src/test/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/ProcessorHelperTest.java
+++ b/src/test/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/ProcessorHelperTest.java
@@ -16,6 +16,7 @@ import org.mockito.Mockito;
 
 import io.neonbee.data.DataContext;
 import io.neonbee.data.internal.DataContextImpl;
+import io.neonbee.entity.EntityWrapper;
 import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.http.impl.HttpServerRequestInternal;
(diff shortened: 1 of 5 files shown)


src/test/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/ProcessorHelperTest.java
```diff
@@ -36,11 +37,12 @@ void transferResponseHint() {
         dataContext.responseData().put(ODATA_FILTER_KEY, Boolean.TRUE);
         dataContext.responseData().put(ODATA_EXPAND_KEY, Boolean.FALSE);
         dataContext.responseData().put(ODATA_COUNT_SIZE_KEY, 42);
-        ProcessorHelper.transferResponseHint(dataContext, routingContext);
+        EntityWrapper result = ProcessorHelper.transferResponseHint(dataContext, routingContext, null);
```
pk-work, line 40: I'd like to use a real object for EntityWrapper, even it is a mock. But this test would ensure that transferResponseHint returns the EntityWrapper you pass into it.

pk-work (changes requested):

pk-work:
Looks good for me, but let's wait for Sebastian.

src/test/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/ProcessorHelperTest.java
dapeng-wang: null is replaced by a mock object

PR #319, opened 2023-05-26
feat: introduce in-memory caching / request coalescing

kristian:
This change introduces a in-memory caching concept for DataVerticle, as well as an option to coalesce same requests sent to a verticle in parallel. The new CachingDataVerticle allows to define an arbitrary cacheKey that is used to store responses received by the DataVerticle in-memory. The next time data with the same cacheKey is requested, neither the required data verticles, nor the original requireData method is being invoked. Instead the data is returned straight from the in-memory cache. This also allows for request coalescing, meaning that if multiple requests are sent to the verticle with the same cacheKey at the same time, the verticle will only do one call to requestData and return the data straight from cache for the other request instead.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,8 @@ dependencies {
 
     implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.7'
     implementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.4.7'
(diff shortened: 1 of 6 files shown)


s4heid (commented):
lgtm! just a few typos

s4heid (approved):

PR #324, opened 2023-05-31
feat: add jsonMaxStringSize configuration

kristian:
This change makes the maximum string size that is parsed by Jackson customizable. This is achieved by a new ConfigurableJsonFactory SPI that Vert.x can use instead of the default JacksonCodec.

With upgrading to Vert.x v4.4.2, Jackson v2.15 introduced a maximum string parsing limit of 5 Million characters, which equals about 5MB in size. This was further increased to 10 Million / 10 MB with Jackson v2.15.1. However with NeonBee we would like to allow users to deal with larger payloads if required.

This change further improves exception handling on the Vert.x event bus, as a negative side-effect of the Jackson change was that event-bus messages reaching the maximum JSON size resulted in uncached exceptions and thus event bus message timing out, which is the worst case scenario. This change now adds specific tests for failures in different lifecycle states of a message (e.g. when being deserialized on the retriever side vs. being deserialized on the message return), as well as handles failing messages much more gracefully and with a descriptive error message.


src/test/java/io/neonbee/internal/json/ConfigurableJsonFactoryClusterTest.java
```diff
@@ -0,0 +1,122 @@
...
+        nb2.getVertx().deployVerticle(testVerticle).map(any -> new DataRequest(testVerticle.getName()))
+                .compose(dr -> DataVerticle.requestData(nb1.getVertx(), dr, new DataContextImpl()))
+                .onComplete(testContext.failing(throwable -> {
+                    assertStringLengthExceptionMessage(throwable);
+                    testContext.completeNow();
+                }));
+    }
+
+    @Test
+    // @Timeout(value = 5, timeUnit = TimeUnit.SECONDS)
```
s4heid, line 97: why is the timeout commented? guess we should comment it in again.

s4heid (changes requested):
that's a good change, thanks kristian. if we could add the new config parameter to the docs I guess we can merge this change.

src/test/java/io/neonbee/internal/json/ConfigurableJsonFactoryClusterTest.java
pk-work: I think we need to remove all timeouts, the timeout is controlled by the environment, since https://github.com/SAP/neonbee/commit/02d22528

src/generated/java/io/neonbee/config/NeonBeeConfigConverter.java
```diff
@@ -40,6 +40,11 @@ static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, NeonBee
                             new io.neonbee.config.HealthConfig((io.vertx.core.json.JsonObject) member.getValue()));
                 }
                 break;
+            case "jsonMaxStringSize":
```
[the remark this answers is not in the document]
pk-work: Done

kristian:
one test failed but lgtm

pk-work (approved):

s4heid (approved):

PR #326, opened 2023-06-02
feat: return 200 (not 204) for PUT requests to entity verticles

dapeng-wang:


docs/usage/verticles/EntityVerticle.md
```diff
@@ -0,0 +1,35 @@
...
+
+A `POST` request will be used to create a new entity. In this case, the `Future<EntityWrapper> createData(DataQuery query, DataContext context)` method of the responsible `EntityVerticle` will be invoked.
+According to the best-practices of REST-API, the response of such a POST-request should
+* contain a `Location`-header with the location of the newly created entity
+* return the entity representation in the response body
+
+To achieve this, a `EntityVertile` should in the `createEntity`-method
+* set the `Location` in the response context data
+```
+context.responseData().put("Location", /io.neonbee.test3.TestService3/TestCars(/unique-id));
```
pk-work, line 13: When we send back a filled EntityWrapper, we have the entity type informatiomn and the unique-id available in the EntityProcessor. Why not simply "generate" the location handler? No need to send also the location header in response data.

docs/usage/verticles/EntityVerticle.md
```diff
@@ -0,0 +1,35 @@
...
+```
+* return a filled `EntityWrapper`
+```
+return Future.succeededFuture(new EntityWrapper(TEST_ENTITY_SET_FQN, createEntity(ENTITY_DATA_1));
+```
+In this case, a 201 status code with the entity JSON representation will be returned in the HTTP-response.
```
pk-work, line 19: the representation can be JSON or XML depending on the request params.

docs/usage/verticles/EntityVerticle.md
dapeng-wang: It depends how the URL convention is in the application, where the EntityVerticle is deployed.

For FORME, a URL like /backend/odata/namespace/type(key) should be returned, so that frontend can really use this header. 

Other application might choose to use other prefixes or no prefix at all.

docs/usage/verticles/EntityVerticle.md
dapeng-wang: Removed JSON to make it format neutral

docs/usage/verticles/EntityVerticle.md
pk-work: I'm not convinced. The URL Convention is configured in the ServerVerticle / Endpoint, on the web-facing node. So the web-facing node should be ideal to "generate" the correct URL.

In my opinion the EntityVerticle is the wrong place to "generate" the location. Let us discuss this in the next committer meeting.

pk-work:
In the last NeonBee Committer Meeting we discussed this PR with @kristian . We came to the following conclusion:

1. If a Response has a location header as kind of a response hint (e.g. response.headers.Location) then use the value of the hint.
2. If no response hint is present, check if entity wrapper contains created entity and build "key" based on the entity. This key is then used as location header value.
3. If not hint and no entity present, don't send a location header

s4heid (approved):
Only found the two typos in the doc. Implementation looks good to me :)

I created https://github.com/SAP/neonbee/issues/374 to tackle the leftovers of Pascals comment. As agreed in the last committer meeting, we split the implementation.

s4heid (approved):

PR #332, opened 2023-06-07
build(deps): exclude caffeine from vertx-infinispan

ada89:
caffine is causing issues during the local development, as gradle is not able to resolve caffeine to a certain version


pk-work:
Ca you explain what's the issue and which version can't be resolved? It seems that caffeine stores its artifacts like other projects on maven central https://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/

dapeng-wang (approved):

PR #342, opened 2023-06-19
feat: expose module registration for Databind codec

pk-work:
This change enables the registration of custom modules for a DatabindCodec. It also improves the selection logic for the deserialization codec.


halber (approved):
LGTM

PR #358, opened 2023-07-18
feat: extend NeonBeeOptions with trust configuration

pk-work:
With this commit it is possible to configure a keystore and a truststore to enable event bus encryption.

The keystore and truststore must be in the PKCS #12 format.


tschuba (approved):

src/main/java/io/neonbee/NeonBee.java
```diff
@@ -314,9 +317,34 @@ static Future<Vertx> newVertx(VertxOptions vertxOptions, NeonBeeOptions options)
...
-            LOGGER.error("Failed to start clustered Vert.x", throwable); // NOPMD slf4j
-        });
+
+        return applyEncryptionOptions(options, vertxOptions.getEventBusOptions())
+                .compose(v -> Vertx.clusteredVertx(vertxOptions)).onFailure(throwable -> {
+                    LOGGER.error("Failed to start clustered Vert.x", throwable); // NOPMD slf4j
+                });
+    }
+
+    @SuppressWarnings("PMD.EmptyIfStmt")
```
halber, line 327: I find the @SuppressWarnings("PMD.EmptyIfStmt") unnecessary. Why not simply return in each case a future?

halber (approved):

PR #386, opened 2023-09-20
feat: add Hooks / Handlers RFC

pk-work:


kristian:
There are currently two stages in each routing chain:

- One for the `ServerVerticle`, that is configured by the `ServerVerticle` configuration and then
- One sub-chain for each `Endpoint` that is registered

As we do allow each `Endpoint` to use a different authentication chain (or the default one), we do register the `AuthHandler` in the endpoint, whilst e.g. the `ErrorHandler` is only configured in the server configuration once for all endpoints, so we register it it the `ServerVerticle` itself.

The existing `ONCE_PER_REQUEST` handler already receives the `RoutingContext`, from which you can derive most the information and also register a end handler using `addEndHandler`. From my point of view it would be important to not introduce duplicate methods, if `ONCE_PER_REQUEST` and `addEndHandler` can maybe already cover some of the additional handlers you suggest. Same is true for the parameters, if you can determine the correlation ID from the ROUTING_CONTEXT parameter that we already pass in, we do not need to pass in duplicate information into the hook.

I don't like the naming scheme of the hooks yet, as the `INCOMING_WEB_REQUEST` prefix makes the name very long and still the most useful one for most scenarios will be the `INCOMING_WEB_REQUEST_AUTHENTICATED` because authentication already happened. So we should reflect that also in the name that this one should stay "the default handler".

SAP AppRouter has a similar extension concept, they call it either a `first` handler (before every other request), in our case I would suggest, we register it in the `ServerVerticle` "globally" and then they provide a `beforeRequestHandler` which is essentially equivalent to our `ONCE_PER_REQUEST` handler. Last they provide a `afterRequestHandler` which can be registered in one of the two `first` / `beforeRequestHandler` which is similar to Vert.x's `addEndHandler` concept.

So from my point of view, it would be reasonable to think about:

1) Do we really need a `INCOMING_WEB_REQUEST_FAILED` handler if we have a `addEndHandler` on the `RoutingContext` (that way you can also correlate between inbound + outbound, because you can register a handler while in the `ONCE_PER_REQUEST`.
2) Shouldn't we just keep `ONCE_PER_REQUEST` and instead of adding a `INCOMING_WEB_REQUEST` add a hook that is registered in the `INCOMING_WEB_REQUEST`` , called `BEFORE_REQUEST` or similar?

pk-work:
Okay, adding an endHandler to the RoutingContext and pass a reference of the RoutingContext into the endHandler to be able to access the the failure in case a request fails, could work. But this would require a "BEFORE_REQUEST" hook to ensure that this endHandler is always added.

```java
rtx.addEndHandler(asyncVoid -> {
  if(rtx.failed() && rtx.statusCode == 401) {
    // write audit log
  }
});
```

Maybe we can discuss the hook name "ONCE_PER_REQUEST". The more I think about it, the less I like it. 

- I think the main issue I have with this name is, that it is misleading. The "BEFORE_REQUEST" hook would be really a "ONCE_PER_REQUEST" hook.
- The "ONCE_PER_REQUEST" hook is called after endpoint specific authentication was done, but before the SubRouter created by an Endpoint is mounted. In my opinion a name in the direction of "REQUEST_PASS_AUTHENTICATION" or "BEFORE_ENDPOINT_ROUTER" would fit better. Such a name would also reveal when the hook is called.

PR #406, opened 2023-10-24
fix: add compare UUID with UUID string

halber:


src/test/java/io/neonbee/endpoint/odatav4/internal/olingo/expression/EntityComparisonTest.java
```diff
@@ -60,4 +69,24 @@ void instanceOfExpectedTypeTest() {
...
+
+    static Stream<Arguments> comparePropertyValuesParameters() {
+        UUID uuid1 = UUID.fromString("1386d8da-4cb4-4fee-97ef-5a24d6a41b34");
+        UUID uuid2 = UUID.fromString("74affcbf-1c66-4903-9cf3-685f511c93b0");
+
+        return Stream.of(
+                Arguments.of(null, uuid1, uuid1, EdmPrimitiveTypeKind.Guid, "ID", 0),
+                Arguments.of(null, uuid1, uuid1.toString(), EdmPrimitiveTypeKind.Guid, "ID", 0),
+                Arguments.of(null, uuid1, uuid2, EdmPrimitiveTypeKind.Guid, "ID", -1),
+                Arguments.of(null, uuid2, uuid1.toString(), EdmPrimitiveTypeKind.Guid, "ID", 1));
```
dapeng-wang, line 81: only UUID instances are compared here. Isn't it better to have the UUID and String comparison covered as well?

dapeng-wang (approved):

PR #440, opened 2024-01-11
BREAKING CHANGE: set Automatic-Module-Name

halber:
Currently, the automatic module name is resolved to `neonbee.core`. This commit changes the automatic module name to `io.neonbee.core`.


kristian (approved):
lgtm

PR #445, opened 2024-01-15
chore: deprecate Java 11, introduce Java 21, fix deprecations

kristian:


src/main/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/CountEntityCollectionProcessor.java
```diff
@@ -102,7 +97,8 @@ public void readEntityCollection(ODataRequest request, ODataResponse response, U
             ContentType responseFormat) {
         List<UriResource> resourceParts = uriInfo.getUriResourceParts();
         if (resourceParts.size() > 2) {
-            throw TOO_MANY_PARTS_EXCEPTION;
+            throw new UnsupportedOperationException(
+                    "Read requests with more than two resource parts are not supported.");
```
pk-work, line 101: What is the benefit of removing the constant?

kristian: According to Error Prone it is a code smell, see [1]

[1] https://errorprone.info/bugpattern/StaticAssignmentOfThrowable

src/main/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/EntityProcessor.java
```diff
@@ -163,7 +158,8 @@ public void init(OData odata, ServiceMetadata serviceMetadata) {
     @Override
     public void readEntity(ODataRequest request, ODataResponse response, UriInfo uriInfo, ContentType responseFormat) {
         if (uriInfo.getUriResourceParts().size() > 2) {
-            throw TOO_MANY_PARTS_EXCEPTION;
+            throw new UnsupportedOperationException(
+                    "Read requests with more than two resource parts are not supported.");
```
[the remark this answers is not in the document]
kristian: Same

src/main/java/io/neonbee/endpoint/odatav4/internal/olingo/processor/CountEntityCollectionProcessor.java
pk-work: Ah cool, I wasn't aware of this. Of course if this produces always the same stack trace it is bad.

gradle/distribution.gradle
```diff
@@ -107,6 +107,7 @@ task testJavadoc(type: Javadoc) {
     classpath += sourceSets.test.compileClasspath
     classpath += sourceSets.main.compileClasspath
     source = sourceSets.test.allJava
+    options.addStringOption('Xdoclint:none', '-quiet')
```
pk-work, line 110: With this option we will ignore JavaDoc warnings, won't we? Why is it necessary?

src/test/java/io/neonbee/internal/json/ConfigurableJsonFactoryTest.java
```diff
@@ -66,65 +50,6 @@ void resetToDefault() {
         ConfigurableJsonFactory.CODEC.setMaxStringLength(DEFAULT_MAX_STRING_LEN);
     }
 
-    @Test
-    @DisplayName("Databind module registration should be ignored if no Databind Codec is used")
-    void testIgnoreModuleRegistrationWithoutDatabindCodec() throws Throwable {
```
pk-work, line 71: Why are these tests no longer necessary?

src/test/java/io/neonbee/internal/verticle/HealthCheckVerticleTest.java
```diff
@@ -60,17 +54,4 @@ void testSharedMap(VertxTestContext testContext) {
                     testContext.completeNow();
                 })));
     }
-
-    @Test
-    @DisplayName("should not register in shared map when non-clustered mode")
```
pk-work, line 65: Why is this test no longer necessary?

pk-work:
Thank you @kristian, that looks like very hard work. Thank you for doing this painstaking work!

halber (commented):
It looks good to me. However, I also don't understand why the tests are no longer necessary.

gradle/distribution.gradle
kristian: The quiet flag was only added for JavaDoc in TEST classes... because it started to complain A LOT

src/test/java/io/neonbee/internal/json/ConfigurableJsonFactoryTest.java
kristian: In regards of the two tests I removed, these where the tests, that wrote into final private variables inside of other classes, which is no longer possible in Java 21. I looked into the tests, essentialyl they just tested a single if (so if true was the method called or not) and because it is testing this if in regards to a static variable of a Vert.x class, there was no other way to test it, instead of overriding that static final field... so I said, becuase of only testing a single if and being the only test that requires overriding finals, it is not worth the effort to keep it.

src/test/java/io/neonbee/internal/verticle/HealthCheckVerticleTest.java
kristian: See above

halber (approved):
LGTM

PR #449, opened 2024-01-17
fead: Allow to set HTTP response values

### LINKED ISSUE #365: [Feature]: Handle content types in RawEndpoint

### The Problem

atm the RawEndpoint only supports text/plain and application/json.

```java
HttpServerResponse response = routingContext.response()//
                                .putHeader("Content-Type",
                                        Optional.ofNullable(context.responseData().get("Content-Type"))
                                                .map(String.class::cast).orElse("application/json"));
                        if (result instanceof JsonObject) {
                            result = ((JsonObject) result).toBuffer();
                        } else if (result instanceof JsonArray) {
                            result = ((JsonArray) result).toBuffer();
                        } else if (!(result instanceof Buffer)) {
                            // TODO add logic here, what kind of data is returned by the data verticle and what kind of
                            // data is ACCEPTed by the client. For now just support JSON and always return
                            // application/json.
                            result = Json.encodeToBuffer(asyncResult.result());
                        } else {
                            // fallback to text/plain, so that the browser tries to display it, instead of downloading
                            // it
                            response.putHeader("Content-Type", "text/plain");
                        }
```

Ignoring the content type causes the user to return an encoded string and create a file from that.

### Desired Solution

Create a concept to handle content types in the RawEndpoint and react to the given request content type headers.

### Alternative Solutions

_No response_

### Additional Context

_No response_

halber:
This commit closes #365.

### DIFF
--- a/src/main/java/io/neonbee/NeonBee.java
+++ b/src/main/java/io/neonbee/NeonBee.java
@@ -47,6 +47,7 @@ import io.neonbee.config.NeonBeeConfig;
 import io.neonbee.config.ServerConfig;
 import io.neonbee.data.DataException;
 import io.neonbee.data.DataQuery;
+import io.neonbee.endpoint.raw.RawResponse;
 import io.neonbee.entity.EntityModelManager;
 import io.neonbee.entity.EntityVerticle;
 import io.neonbee.entity.EntityWrapper;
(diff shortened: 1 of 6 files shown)


halber:
see https://github.com/SAP/neonbee/pull/450

PR #464, opened 2024-01-31
feat: allow to set HTTP response values

halber:
All HTTP headers from the `DataContext` response data `RESPONSE_HEADERS_HINT` are set in the HTTP response.

### DIFF
--- a/src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
+++ b/src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
@@ -86,7 +86,6 @@ public class RawEndpoint implements Endpoint {
 
     @VisibleForTesting
     static class RawHandler implements Handler<RoutingContext> {
-        private static final String CONTENT_TYPE = "Content-Type";
 
         private final boolean exposeHiddenVerticles;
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
```diff
@@ -88,6 +88,8 @@ public Future<Router> createEndpointRouter(Vertx vertx, String basePath, JsonObj
     static class RawHandler implements Handler<RoutingContext> {
         private static final String CONTENT_TYPE = "Content-Type";
 
+        private static final String RESPONSE_HEADERS = "RESPONSE_HEADERS";
```
kristian, line 91: I think it makes sense to make this public, because consumers need to set the value in their verticles and they should use the constant instead of had coding the value. Also I would move it to Endpoint.java, as it is relevant for all endpoints. Also I would call it "RESPONSE_HEADERS_HINT", as we agreed that everything that the endpoints evaluate we treat as a "hint" towards the endpoints.

src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
```diff
@@ -88,6 +88,8 @@ public Future<Router> createEndpointRouter(Vertx vertx, String basePath, JsonObj
     static class RawHandler implements Handler<RoutingContext> {
         private static final String CONTENT_TYPE = "Content-Type";
```
kristian, line 89: I know it is not part of the commit, but similar to my lower comment, I would suggest we do the same, rename this to "CONTENT_TYPE_HINT" and move it to Endpoint.java, also we should use the same structure for the value.

ada89 (approved):
checked Kristian's comments on the other PR and LGTM

PR #465, opened 2024-01-31
fead: allow to set HTTP response values

halber:


halber:
It was decided that we would set the values in the [DataContext#responseData](https://github.com/SAP/neonbee/blob/8abf1f1cc4213f821fbedd4292e41f8522e56660/src/main/java/io/neonbee/data/DataContext.java#L106C25-L106C37) for the HTTP response, which I did in the following PR https://github.com/SAP/neonbee/pull/464 .

PR #467, opened 2024-02-08
Feat/http response headers nb0.31

halber:
feat: allow to set HTTP response values

All HTTP headers from the `DataContext` response data
`RESPONSE_HEADERS` are set in the HTTP response.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ plugins {
 }
 
 group = 'io.neonbee'
-version = '0.31.1'
+version = '0.31.2-SNAPSHOT'
 mainClassName = 'io.neonbee.Launcher'
 archivesBaseName = 'neonbee-core'
 sourceCompatibility = 11
--- a/src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
+++ b/src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
@@ -86,7 +86,6 @@ public class RawEndpoint implements Endpoint {
 
     @VisibleForTesting
     static class RawHandler implements Handler<RoutingContext> {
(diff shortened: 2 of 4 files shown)


src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
```diff
@@ -178,6 +180,12 @@ queryParameterMap, multiMapToMap(request.headers()), routingContext.body().buffe
                         }
 
                         HttpServerResponse response = routingContext.response();
+
+                        Optional.ofNullable(context.responseData().get(RESPONSE_HEADERS))
```
ada89, line 184: could `context.responseData()` be null?

build.gradle
```diff
@@ -21,7 +21,7 @@ plugins {
 }
 
 group = 'io.neonbee'
-version = '0.31.1'
+version = '0.31.2-SNAPSHOT'
```
ada89, line 24: should it be snapshot? not sure what's the release process like in neonbee

src/main/java/io/neonbee/endpoint/raw/RawEndpoint.java
halber: In this case, it's not possible because we're creating the context (DataContextImpl) object, and its implementation of responseData always returns a map.
But I could also write this null-save. I'll change that.

build.gradle
halber: Normally this is done automatically, that the version is set to SNAPSHOT after NeonBee is released, but because we released NeonBee 0.31.1 "manually", I had to change this.

ada89 (approved):
LGTM

PR #502, opened 2024-04-17
fix: odata guid type casting issue

halber:
This commit resolves an error encountered when applying $filter with GUID values in an OData endpoint.


carlspring (approved):
LGTM!

PR #525, opened 2024-06-06
fix: mitigate problem with ChainAuthHandler

halber:
A problem with the ChainAuthHandler was introduced in vert.x version 4.5.6. To mitigate the issue, we do not use the vert.x ChainAuthHandler when only one AuthenticationHandler is used. The problem still exists when more than one AuthenticationHandler is used.

https://github.com/vert-x3/vertx-web/issues/2611


carlspring (approved):
LGTM! :)

PR #530, opened 2024-06-24
feat: handling for failed deployments during boot

kristian:
Previously any failure during verticle deployment during boot could lead to the bootstrap of NeonBee failing / undeployment of all verticles. This commit introduces a new "bootDeploymentHandling" option in the NeonBeeConfig, which lets the user specify how to deal with failures during verticles / module deployments during boot. The change is compatible with how it is handled currently (FAIL_ON_ERROR), however also provides options to undeploy any failing modules (UNDEPLOY_FAILING) or even keeping modules that could only be partially deployed (KEEP_PARTIAL).


halber (dismissed):
LGTM

PR #547, opened 2024-08-06
fix: null check

ikalachy:
The following exception potentially breaks the hazelcast cluster and we need to restart the cluster + approuter.

```
java.lang.NullPointerException: Cannot invoke "io.vertx.core.json.JsonArray.copy()" because the return value of "java.util.Map.remove(Object)" is null,
at io.neonbee.internal.cluster.entity.ClusterEntityRegistry.lambda$unregisterNode$1(ClusterEntityRegistry.java:127),
```


carlspring (approved):
LGTM! :)

halber (approved):

carlspring (approved):
LGM! :)

PR #580, opened 2024-10-24
fix: entity unregister node

halber:
When the `unregisterNode` method is called for another NeonBee node, and the executing node has an `EntityVerticle` that was also registered by the other node, the `EntityVerticle` was incorrectly unregistered for the executing node. This fix ensures that the `EntityVerticle` remains registered for the executing node.

### DIFF
--- a/src/test/java/io/neonbee/internal/cluster/entity/UnregisterEntitiesTest.java
+++ b/src/test/java/io/neonbee/internal/cluster/entity/UnregisterEntitiesTest.java
@@ -4,56 +4,87 @@ import static com.google.common.truth.Truth.assertThat;
 import static io.neonbee.NeonBeeInstanceConfiguration.ClusterManager.HAZELCAST;
 import static io.neonbee.NeonBeeInstanceConfiguration.ClusterManager.INFINISPAN;
 import static io.neonbee.NeonBeeProfile.WEB;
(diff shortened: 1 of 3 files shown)


src/test/java/io/neonbee/internal/cluster/entity/UnregisterEntitiesTest.java
```diff
@@ -65,42 +96,372 @@ private void testUnregisteringEntities(NeonBee web, VertxTestContext testContext
...
+        vertx.setPeriodic(100, id -> { // Check every 100 ms
+            if (cluserFormed.getAsBoolean()) {
+                vertx.cancelTimer(id);
+                LOGGER.info("The cluster has formed after {} ms.", System.currentTimeMillis() - startTime);
+                promise.complete();
+            } else if (timer.isComplete()) { // fail after ~60 seconds
+                vertx.cancelTimer(id);
+                promise.fail(
+                        new IllegalStateException("Cluster did not form in time. The waited for "
+                                + (System.currentTimeMillis() - startTime) + " ms."));
```
carlspring, line 197:
```suggestion
                        new IllegalStateException("Cluster did not form in time. (Waited for "
                                + (System.currentTimeMillis() - startTime) + " ms)."));
```

carlspring (approved):
LGTM! :)

carlspring (approved):

PR #584, opened 2024-11-04
Refactor/registry

halber:


src/main/java/io/neonbee/internal/SharedRegistry.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Register a value in the async shared map of {@link NeonBee} by key.
+     *
+     * @param sharedMapKey the shared map key
+     * @param value        the value to register
+     * @return the future
+     */
+    @Override
+    public Future<Void> register(String sharedMapKey, T value) {
+        logger.info("register value: \"{}\" in shared map: \"{}\"", sharedMapKey, value);
```
carlspring, line 58: I'm not a big fan of this `if (logger.isInfoEnabled())` approach, (but this just my own opinion). However, as I can see you're following it, to me it would make more sense to be consistent and have it across the rest of your changes as well. (Both here and below).

src/main/java/io/neonbee/internal/SharedRegistry.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Unregister the value in the {@link NeonBee} async shared map from the sharedMapKey.
+     *
+     * @param sharedMapKey the shared map key
+     * @param value        the value to unregister
+     * @return the future
+     */
+    @Override
+    public Future<Void> unregister(String sharedMapKey, T value) {
+        logger.debug("unregister value: \"{}\" from shared map: \"{}\"", sharedMapKey, value);
```
carlspring, line 86: As above.

carlspring (dismissed):
LGTM!
Thanks for putting together some tests as well.

src/main/java/io/neonbee/internal/SharedRegistry.java
halber: Normally PMD would complain here. But as it does not recognise our logger, PMD ‘overlooks’ this. In general, however, I'm not a fan of this either and would rather have a logger that evaluates lambda expressions as parameters.

carlspring (approved):
LGTM! :)

PR #585, opened 2024-11-11
Feat/clean up cluster information

halber:


halber:
The feature is currently not needed and therefore not merged

PR #591, opened 2025-01-24
fix: exclude deps

ikalachy:


halber (commented):
Doesn't it make more senes to exclude this dependencies for all dependencies?
e.g.:
```
configurations.all {
    exclude group: 'org.infinispan', module: 'infinispan-core'
}
```

halber (approved):
LGTM

PR #599, opened 2025-02-28
fix: prevent potential memory leak

halber:
ThreadLocal variables should be explicitly cleaned up when no longer used to prevent memory leaks, especially in environments with thread pools.

- Called `remove()` on `THREAD_LOCAL_ODATA` to ensure proper cleanup.


carlspring (approved):
LGTM! :)

PR #600, opened 2025-03-03
fix(ci): update actions/cache to v4 to prevent deprecation issues

halber:
This commit updates the GitHub Actions workflow to use `actions/cache@v4` instead of the deprecated `actions/cache@v2`. Failure was caused by GitHub's deprecation of v1/v2, effective December 5, 2024. See: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-break


carlspring (approved):
LGTM! :)

PR #608, opened 2025-06-20
fix: refactoring classpath scanning to 

ikalachy:
Refactored scanForAnnotation for:

- Avoiding unnecessary intermediate objects
- Using efficient iteration instead of streams where possible
- Handling I/O safely
- Reducing memory pressure by avoiding distinct().toList() on potentially large collections

Also:

- Refactored mutableCopyOf() methods for List, Set, Map, and generic Collection to:
- Reduce stream overhead.
- Use proper initial capacities to minimize reallocation.
- More efficient multiMapToMap.

image


carlspring (approved):
LGTM! :)

PR #692, opened 2026-05-21
fix: skips registration when collectClusteredResults false.

ikalachy:
The change at HealthCheckVerticle.java:47-48 now skips the distributed lock registration when collectClusteredResults is false. 
Since no node will read the shared map in that configuration, there's no reason to write to it — and the startup lock contention is mitigated


carlspring (approved):
LGTM!

PR #733, opened 2026-06-09
feat: offload large payload serialization to worker thread

ikalachy:
When a DataVerticle replies with a JsonArray or JsonObject over the clustered event bus, the Jackson serialization in encodeToWire can block the event loop for large payloads. This change offloads all JSON reply serialization to a worker thread via executeBlocking.

A configurable threshold (eventBusLargePayloadThreshold) controls when a warning is logged about large payloads. It defaults to the jsonMaxStringSize value if set, aligning both inbound parsing limits and outbound payload monitoring to the same configured size.


carlspring (approved):
LGTM! :)

PR #734, opened 2026-06-10
fix: prevent StackOverflowError in CollectionHelper 

ikalachy:
circular references

Add cycle detection using IdentityHashMap to track already-copied objects during recursive deep copy. When a circular reference is encountered, the previously created copy is returned instead of recursing infinitely.


carlspring (dismissed):
LGTM.

PR #737, opened 2026-06-12
fix: resolve race condition in WriteSafeRegistry lock release

ikalachy:
Use Future.andThen() instead of onComplete() for lock release to ensure the distributed lock is fully released before the next queued operation starts acquiring it. Also skip redundant put() in SharedRegistry when the value already exists, and add polling helper to cluster test for CI stability.


carlspring (approved):
LGTM!
Also, thanks for the code cleanup.

PR #749, opened 2026-06-23
fix(test): replace fragile instance-identity check in EntityModelLoadTest

ikalachy:
The previous assertion relied on CdsModelReader's internal Caffeine weak-reference cache returning the same CdsModel instance for two separate readCsnModel calls on the same file. Under GC pressure on CI the weak reference is collected between the two calls, producing different instances and a spurious test failure.

Replace with a structural assertion on the qualified names of services and entities, which is stable, meaningful, and independent of caching behaviour.


carlspring (approved):
LGTM, thanks!

PR #754, opened 2026-06-26
fix: skip OData services with null EntityContainer to prevent NPE

ikalachy:
Services whose EDMX metadata has no EntityContainer are now filtered out during endpoint registration instead of causing a NullPointerException. Adds a warning log when an associated model path is missing and a null filter before collecting edmx models. Test coverage added via EmptyContainerService.


ikalachy:
+ one more test fix from https://github.com/SAP/neonbee/pull/749

radrt (approved):

PR #800, opened 2026-08-24
feat: improve publishing workflow

carlspring:
* Make it possible to trigger the workflow manually.
* Rollback automatically, if either of the publishing steps fail.


ikalachy:
could we also update https://github.com/SAP/neonbee/blob/main/docs/internal/howto_release_neonbee.md

